PR #1, opened 2019-09-24
Remove ./iostat and ./sar dirs

atheurer:
-Directories like this under subprojects/tools|benchmarks
 should not be present.  These directories can be built
 on demand at crucible install/sync.  Another config file
 will be maintained to list what subprojects should be
 bundled with crucible.  We only need ./subprojects and
 ./subprojects/tools|benchmarks because those are considered
 well known locations within crucible project to git-clone
 various subprojects.


k-rister:
This change looks fine to me, but I do have a question.  Are we going to use actual git subprojects or are we just going to have the install/setup script take care of cloning the various projects into the directory structure?

atheurer:
I am really not sure on the git subproject.  I have never really used it myself.  If we do use submodules, it will basically take care of tracking these projects for us with .gitmodules in each of the subprojects, subprojects/benchmarks, and subprojects/tools directories.  But I think that will also control what branches/commit levels each of those projects are sync'd to.  We may not actually want that if we want to have devs/users pick what branches they use.

atheurer:
And this PR is IMO more about making sure the process works for migrating changes from my personal repo to the project repo.

k-rister:
The submodules thing is pretty powerful as I recall, I think by default it just uses the HEAD of the master branch but I am pretty sure you can specify alternative branches and/or commits.  I'm pretty sure we did that in trafficgen for MoonGen.  I agree as to the purpose of this PR so I filed an approval, mostly I was just looking to provide some documented discussion on these types of questions, but I guess that is probably more suitable for Issues.

PR #6, opened 2019-11-04
add infrastructure for updating project repositories

k-rister:


k-rister:
I think I'm going to do a small rework of this code.  I'm trying to anticipate a problem that I could see happening and on further reflection I think I see a better solution than what I did here.  The problem being is that when updating crucible, if the update includes changes to the update script(s) then there is potential for breakage since shell doesn't handle changes to it's scripts happening while they are running very well.  I will probably ask for another review when I get the new approach written.

PR #7, opened 2019-11-07
logger infrastructure

k-rister:


k-rister:
@atheurer 
The commits are messy right now but if you are happy with the code I'll squash the commits and come up with a better commit message.

k-rister:
Side note, there is some perl code that is duplicated in two separate scripts for formatting the timestamps and printing the log messages.  If we come up with a place to put library code this would be a good place to start consolidating some code.

k-rister:
> Also forgot to add, a squash would be nice, and can you add the perl header I have in multiplex with the vim options?

Yep, I plan on squashing, just figured I would wait until I was sure it was done to do that in case I wanted to look at how I got here.

I'll add the header, not a problem.

k-rister:
> It looks fine, but I am wondering, for future PR's, what would be the best way to test this before we approve something. I suppose eventually we will have a workflow where the PR is against a test branch, and I can merge, then switch to the test branch to test myself. However, if it does not work, then we have to back it out of the test tree, or just keep applying commits to get it working. Ideally it would be really nice if this repo (incubator/crucible) could just temporarily duplicate your branch (dev-kmr), and so all I would have to do is switch branches on my local install, run some tests, and once it seems alright, then merge to the next branch in the workflow (master today, eventually something else).

I think you should be able to pull the remote branch into your local repo without too much problem for testing purposes, I'll figure out the workflow to share and then maybe we can go about automating it.

Also, I was tossing around the idea of working on a TravisCI testing infrastrucuture, but hopefully with much less (or none) of the mocking stuff we are used to.

k-rister:
> I think you should be able to pull the remote branch into your local repo without too much problem for testing purposes, I'll figure out the workflow to share and then maybe we can go about automating it.

git remote add <remote-name> <repo-url>  # <remote-name> is a local identifier of your choice
git fetch --all
git checkout --track <remote-name>/<branch-name>

atheurer:
OK, that process worked well for me.  This makes it quite easy to test before merging.

atheurer:
Ahh, this is fantastic.  Merge, then 'git checkout master', then 'crucible update' and you are back on "GA" code.  This is infinitely better than the process needed for our previous tooling.  At some point we should probably write up how developers and admins do this.

k-rister:
This afternoon I started working on "crucible repo <command>" that can be used to check on repo status and eventually do the heavy lifting of switching repos and the testing procedure outlined here.

PR #15, opened 2020-01-16
logger: convert to using SQLite as the backing store for the log

k-rister:
- This has many benefits including, but not limited too:

  1. Support for concurrent logging since SQLite supports multi
     process access:

     https://www.sqlite.org/faq.html#q5

  2. Easily queried data for find specific information

### DIFF
--- a/bin/_crucible_completions
+++ b/bin/_crucible_completions
@@ -23,7 +23,7 @@ _crucible_completions() {
                 COMPREPLY=($(compgen -W "log repo update gen-iterations" -- "${COMP_WORDS[2]}"))
                 ;;
 	    log)
-		COMPREPLY=($(compgen -W "clear info view" -- "${COMP_WORDS[2]}"))
+		COMPREPLY=($(compgen -W "clear info init insert view" -- "${COMP_WORDS[2]}"))
 		;;
 	    repo)
 		COMPREPLY=($(compgen -W "status" -- "${COMP_WORDS[2]}"))
--- a/bin/_logger
+++ b/bin/_logger
@@ -8,15 +8,21 @@ use warnings;
 
 use Time::HiRes qw(time);
 use IO::Select;
+use DBI;
 
 # disable STDOUT output buffering
 $|++;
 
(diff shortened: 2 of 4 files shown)


k-rister:
Let's not merge this quite yet, but I wanted you to have the chance to look over it...

k-rister:
Can you test this out and see what you think?  If you think it's done I'll squash the commits together before merging.  I don't really want to commit them individually since the DB schema changes between them and it would potentially break if you jumped from one commit to another.

atheurer:
Sure, as soon as I remember how to map your branch to my git repo

atheurer:
So one thing that has occurred to me, not just with your commit but all possible commits from crucible and any project it uses, is that we probably need to check for install dependencies at least when someone runs ./bin/install.  In this case I did not have the DBI.pm, and actually could not control-c out of "crucible log info" which was odd.  I don't know if we should assume a certain type of package manager like yum/dnf.  And perhaps things like our perl programs can abort cleanly of the perl module needed is not found.

k-rister:
> So one thing that has occurred to me, not just with your commit but all possible commits from crucible and any project it uses, is that we probably need to check for install dependencies at least when someone runs ./bin/install. In this case I did not have the DBI.pm, and actually could not control-c out of "crucible log info" which was odd. I don't know if we should assume a certain type of package manager like yum/dnf. And perhaps things like our perl programs can abort cleanly of the perl module needed is not found.

So the whole console redirection thing has some nasty side affects like the CTRL-C thing you encountered.  I can probably clean this up more with some effort but it's not been something I focused on yet.

As for dependencies, I agree that needs some work.  I have kind of been ignoring it for the time being because I figured we would take care of that when we start building our containers with all of those dependencies.  Sorry I forgot to mention the DBI.pm thing.

k-rister:
> Functionally things seem fine. I can't say I have gone through all the code and understand all of it, but I am fine with merging. At some point later I'll dig through the code and make sure I understand how all this works.

I thought of something that I need to fix but once I get that done I'll squash this and merge it.

PR #22, opened 2020-03-02
bin/update: CRUCIBLE_HOME must be defined

atheurer:
-if not the check for a directory actually passes


k-rister:
Did you try to invoke update directly rather than through the crucible script?  Wondering how you hit this...

PR #23, opened 2020-03-10
Fixes to allow install in a container

atheurer:
-'sudo' may not be available and root is typically used anyway in a container
-'which' may not be available so don't use it, call the bin with --version to check if it is installed

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -10,7 +10,7 @@ function exit_error() {
 # Check for dependencies
 missing_bins=""
 for bin in sed git; do
-    which $bin >/dev/null 2>&1 || missing_bins="$missing_bins $bin"
+    $bin --version >/dev/null 2>&1 || missing_bins="$missing_bins $bin"
 done
 if [ ! -z "$missing_bins" ]; then
     exit_error "Could not find the following, please install and try again: $missing_bins"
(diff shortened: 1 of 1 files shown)


bin/install
```diff
@@ -10,7 +10,7 @@ function exit_error() {
 # Check for dependencies
 missing_bins=""
 for bin in sed git; do
-    which $bin >/dev/null 2>&1 || missing_bins="$missing_bins $bin"
+    $bin --version >/dev/null 2>&1 || missing_bins="$missing_bins $bin"
```
k-rister: Is this a reliable way of checking?  I explicitly listed these as requirements in my example controller userenv:

https://github.com/perftool-incubator/crucible/blob/master/workshop/controller.json

atheurer:
I don't know if the check is a completely reliable.
At least initially I was not going to use workshop to build the container.  I want to try coming up with a dockerfile or whatever github supports because I want to be able to have github build it for us when we trigger a new release.

k-rister:
https://github.com/perftool-incubator/crucible/actions/new

https://github.com/actions/starter-workflows/blob/master/ci/docker-image.yml

bin/install
atheurer: What is "perl-Thread-Queue" in the requirements?

k-rister: It's a thread safe queue library I typically use in my threaded perl programs.  It's used in bin/_logger.

PR #33, opened 2020-04-28
Convert all tabs to spaces

atheurer:
-We should not be using a mix of tabs and spaces

### DIFF
--- a/bin/_logger
+++ b/bin/_logger
@@ -42,71 +42,71 @@ if (! $log_dbh) {
     my $rv = $log_dbh->do("SELECT timestamp FROM db_state;") or $log_open_error = 1;
 
     if ($log_open_error) {
-	printf("ERROR: The SQLite log DB '%s' does not appear to be initialized!\n", $log_filename);
-	exit(1);
+        printf("ERROR: The SQLite log DB '%s' does not appear to be initialized!\n", $log_filename);
+        exit(1);
     } else {
-	if ($rv < 0) {
-	    print $DBI::errstr;
-	    exit(1);
-	}
+        if ($rv < 0) {
+            print $DBI::errstr;
+            exit(1);
+        }
 
-	$rv = $log_dbh->do("UPDATE db_state SET timestamp=$start_timestamp;");
-	if ($rv < 0) {
-	    print $DBI::errstr;
-	    exit(1);
-	}
+        $rv = $log_dbh->do("UPDATE db_state SET timestamp=$start_timestamp;");
+        if ($rv < 0) {
+            print $DBI::errstr;
+            exit(1);
+        }
     }
 
     my $sth = $log_dbh->prepare("SELECT id FROM sources WHERE source=" . $log_dbh->quote($source) . ";");
     $rv = $sth->execute() or die $DBI::errstr;
     if ($rv < 0) {
-	print $DBI::errstr;
-	exit(1);
+        print $DBI::errstr;
+        exit(1);
     }
 
     my $source_id;
     my $rows = 0;
     while (my $row = $sth->fetchrow_hashref()) {
-	$source_id = $row->{'id'};
-	$rows += 1;
+        $source_id = $row->{'id'};
+        $rows += 1;
     }
 
     if ($rows == 0) {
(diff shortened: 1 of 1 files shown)


atheurer:
I am pretty sure it is your fault :)

Your editor has decided to use a mix of tabs and spaces.  The header for vi will not automatically fix old tabs, only adjust new indents, unless you run "retab" every time.  Also, your editor decided to assume single tab is for *two* indents which would make running "retab" completely screw up the indentation.

k-rister:
> I am pretty sure it is your fault :)

I know...I was trying to deflect :)

> 
> Your editor has decided to use a mix of tabs and spaces. The header for vi will not automatically fix old tabs, only adjust new indents, unless you run "retab" every time. Also, your editor decided to assume single tab is for _two_ indents which would make running "retab" completely screw up the indentation.

I agree that the header doesn't fix things, but if we don't add the header when we fix things then it will just happen again.

atheurer:
Yes, we either need either your editor to use non-dumb defaults or get the right mode line for emacs in these files.  You are the emacs expert, not me :)  From what I can google, we need something like:

<pre>
# -*- Mode: C; indent-tabs-mode: nil; c-basic-offset: 4; tab-width: 4 -*-
</pre>

That's for c code.  I imagine it will be slightly different for bash, python, and perl.

k-rister:
Sorry, I thought it was obvious that we had available examples but then I remembered that was in another repo.  From workshop.pl:

```
# -*- mode: perl; indent-tabs-mode: nil; perl-indent-level: 4 -*-
# vim: autoindent tabstop=4 shiftwidth=4 expandtab softtabstop=4 filetype=perl
```

atheurer:
OK, I am going to close this PR and open another to fix all of the files

atheurer:
I assume "bash" can be substituted for "perl" for our bash files...

k-rister:
> I assume "bash" can be substituted for "perl" for our bash files...

I don't think that is true.  If you notice that there is a Perl specific name `perl-indent-level` so other languages might have something different.  I think I recall having to find that to adapt your existing Bash lines from some other file.

atheurer:
ok, do have modeline for bash then?

k-rister:
Apprently not:

```
$ grep -R -- "-\*-" *
bench-fio/fio-post-process:## -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
bench-uperf/uperf-post-process:## -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
crucible/bin/log:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
crucible/bin/_logger:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
junk-drawer/bin/json-validator.pl:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
multiplex/multiplex:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
parking-garage/parking-garage.sh:# -*- mode: sh; indent-tabs-mode: nil; perl-indent-level: 4 -*-
parking-garage/parking-garage.sh~:# -*- mode: bash; indent-tabs-mode: nil; perl-indent-level: 4 -*-
rickshaw/rickshaw-post-process:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
rickshaw/client-server-script:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
rickshaw/rickshaw-run:# -*- mode: perl; indent-tabs-mode: t; perl-indent-level: 4 -*-
tool-proc-watch/bin/watch-sched-debug~:# -*- mode: perl; indent-tabs-mode: nil; perl-indent-level: 4 -*-
tool-proc-watch/bin/watch-sched-debug:# -*- mode: perl; indent-tabs-mode: nil; perl-indent-level: 4 -*-
workshop/workshop.pl:# -*- mode: perl; indent-tabs-mode: nil; perl-indent-level: 4 -*-

$ grep -R -- "-\*-" *|grep -v perl
```

atheurer:
I found some modelines in bash scripts from subprojects, but they still had perl stuff in them.  I am going to try instead:

<pre>-*- mode: sh; indent-tabs-mode: nil; sh-basic-offset: 4 -*-</pre>

PR #41, opened 2020-05-13
More support for using containers including the logger

atheurer:
-container usage is configurable in /etc/sysconfig/crucible
-logging can also be disabled in /etc/sysconfig/crucible
-the 'log debug' command still needs to be implemented when using a container


k-rister:
Oops, I didn't really mean to merge that for you.  I hope that is what you wanted...

atheurer:
fine by me

PR #46, opened 2020-05-15
bin/_logger: DB transaction improvements

k-rister:
- If the DB logger thread goes "idle" and there are outstanding SQL
  INSERT commands in an uncommited transaction, COMMIT the transaction
  intead of sleeping.

- This attempts to ensure that previously performed SQL INSERTs get
  committed ASAP, guarding against loss of log entries in cases such
  as the DB connection failing or the process being uncleanly
  terminated prior to the session ending.

- This change aims to reduce the overhead of the transaction COMMIT by
  only performing it in the "idle" path.

### DIFF
--- a/bin/_logger
+++ b/bin/_logger
@@ -276,6 +276,7 @@ sub db_logger {
 
     my $log_entry;
     my $db_connected = 0;
+    my $rv;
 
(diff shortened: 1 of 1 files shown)


k-rister:
@atheurer I think this is good idea, but I'm not sure of the best way to test/confirm that it is behaving as desired.  Thoughts?

PR #68, opened 2020-09-02
on podman 2.0.5 --privileged and --add-cap=all are mutually exclusive

mrbojangles3:


atheurer:
Tested with F30 and RHEL8 controllers, works on both.

PR #83, opened 2020-11-19
bin/_main: update "crucible run" arg handling

k-rister:
- Rework the argument handler so that parameters that rickshaw-run
  does not understand can be removed from the argument list.

- Create ${run_dir}/config/[bench|tool]-params.json so that these
  files are included in the result directory and use them as the
  values passed to rickshaw-run.

- Optionally allow the user to specify --mv-params (multi-value
  parameters) instead of --bench-params.  If the user provides
  --mv-params then "crucible run" will call multiplex to generate
  --bench-params for the user.  If --mv-params is used it's input is
  also stored as ${run_dir}/config/mv-params.json for inclusion in the
  result.

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -76,27 +76,42 @@ elif [ "${1}" == "run" ]; then
     shift
     base_run_dir=""
     bench_params="bench-params.json"
-    default_bench_params=1
     tool_params="tools-params.json"
-    default_tool_params=1
-    args=("$@")
-    arg_num=0
-    # do some arg processing without modifying $@
-    while [ $arg_num -lt ${#args[@]} ]; do
-        echo "arg $arg_num: ${args[$arg_num]}"
-        if [ ${args[$arg_num]} == "--tags" ]; then
-            # if --tags is used, embedd this in the $base_run_dir
-            base_run_dir="--${args[$arg_num+1]}"
-        elif [ ${args[$arg_num]} == "--bench-params" ]; then
-            # note that --bench-params was provided so skip default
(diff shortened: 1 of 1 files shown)


bin/_main
```diff
@@ -111,11 +126,26 @@ elif [ "${1}" == "run" ]; then
...
+            echo "The multi-value params file you specified with --mv-params (${mv_params}) does not exist!"
+            exit 1
+        else
+            mv_params_run_dir=${base_run_dir}/config/mv-params.json
+            cp -v ${mv_params} ${mv_params_run_dir}
+            bench_params_run_dir=${base_run_dir}/config/bench-params.json
+
+            echo "Generating --bench-params from --mv-params..."
+            echo "$podman_run -i --name crucible-multiplex "${container_common_args[@]}" $CRUCIBLE_CONTAINER_IMAGE /opt/crucible/subprojects/core/multiplex/multiplex.py --input ${mv_params_run_dir} > ${bench_params_run_dir}"
+            $podman_run -i --name crucible-multiplex "${container_common_args[@]}" $CRUCIBLE_CONTAINER_IMAGE /opt/crucible/subprojects/core/multiplex/multiplex.py --input ${mv_params_run_dir} > ${bench_params_run_dir}
```
atheurer, line 144: Did you rewrite mulitplex in python?

atheurer (approved):

bin/_main
k-rister: Yes.  I guess I now know why you never reviewed my PR...

PR #88, opened 2020-11-25
change tools to tool, singlar

mrbojangles3:
I think "tool" is convention. Before this change I was getting an error on the call to `cp` on line 160.


k-rister (approved):
Thanks for fixing this.  Whenever I mess with this area of code I keep typing the wrong way...

PR #94, opened 2020-12-08
add Perl support for xz compression

k-rister:


atheurer (commented):
Is this primarily for rickshaw?

k-rister:
At the moment yes, but I was also assuming that as we get into this deeper we will be using this more pervasively so it might be consumed by other stuff as well.

atheurer (approved):

bin/_main
```diff
@@ -206,7 +204,7 @@ elif [ "${1}" == "run" ]; then
         echo "...appears to be running"
     else
         echo "...not present, starting a container for it:"
-	firewall-cmd --zone=public --add-port=6379/tcp
+        firewall-cmd --zone=public --add-port=6379/tcp
```
atheurer, line 207: Did I have this indent wrong or is this your editor?

atheurer (approved):

bin/_main
k-rister: Well, normally I would assume it is me but for this particular line I can't see how it would be since I don't think I ever touched this particular line.

I still see my editor do some dumb things from time to time while making changes so in the course of fixing those I fixed a few more that my search found such as this line.

I don't understand why my editor is still doing the wrong thing occasionally, I thought the header that you/we added should have taken care of that.  That said, I'm paying a lot closer attention to it now.

PR #96, opened 2020-12-10
add the Apache License 2.0

k-rister:


k-rister:
@atheurer 
This should be here, right?

PR #100, opened 2021-01-05
Update workshop json to add fedora33, centos stream8

atheurer:
-and include bind-utils and iproute, which we need for
 rickshaw

### DIFF
--- a/workshop/controller-workshop.json
+++ b/workshop/controller-workshop.json
@@ -1,376 +1,427 @@
 {
-    "workshop": {
-        "schema": {
-            "version": "2020.03.02"
-        }
+  "workshop": {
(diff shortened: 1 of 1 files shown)


atheurer:
Prior to this commit, fedora33 userenv was built and tested, first as quay.io/crucible/controller:testing, then as quay.io/crucible/controller:latest

The controller workshop file also contains info for a "stream8" (centOS stream8) and opensuse userenv.  Neither of these are fully tested, and in fact opensuse will require a change to how redis is started (not /usr/bin/redis, but just "redis").

There is no real reason to have multiple controller images based on different userenvs, other than to prove that it works.

PR #111, opened 2021-02-25
Crucible standalone installation script

rafaelfolco:
Add crucible standalone install script with a few improvements.
The script depends on the authentication file to be downloaded
separately from secured reporsitory.


atheurer:
Thank you for the PR.  This is mostly fine, but there are a couple issues:

1. I don't think we want a RH specific install script in our upstream repo.  If we want this for general use, then we should just call this something like 'crucible-install.sh'.
2. While it is good that the authorization credentials are not present, we also do not intend to use the same CRUCIBLE_CLIENT_SERVER_REPO publicly.  This is actually another step that a user must take, to set up their own repo (which of course we have not documented yet).  So, I think it would be best to have another option (like --client-server-registry) to specify the registry location.

k-rister:
> Thank you for the PR. This is mostly fine, but there are a couple issues:
> 
> 1. I don't think we want a RH specific install script in our upstream repo.  If we want this for general use, then we should just call this something like 'crucible-install.sh'.
> 2. While it is good that the authorization credentials are not present, we also do not intend to use the same CRUCIBLE_CLIENT_SERVER_REPO publicly.  This is actually another step that a user must take, to set up their own repo (which of course we have not documented yet).  So, I think it would be best to have another option (like --client-server-registry) to specify the registry location.

@rafaelfolco Oops, I forgot that this part was in there.  I agree with Andrew.

atheurer:
This comment is only somewhat related to the installer, but regarding the repo for client-server, we could consider a model where our (RH's) repo is still public, people know about it, and they can pull from it, but they cannot write to it, so they still need a personal repo.  The purpose of this would be to allow these users to search for a matching container image from our repo, and if there is a match, they can pull and use.  If there is not a match, they must then check their personal repo for a match or build/push their image to their personal repo.

If we did this, and users used the default userenv and default set of tools, most likely the images they need to run a benchmark would already be in our repo, and they can just pull and not build.  Also, once we support more incremental builds of the image (builds with partial requirements), there will be more matches.

rafaelfolco:
Thanks for reviewing. I'll address the issues mentioned above.

k-rister:
I wonder if we shouldn't move/rename bin/install as part of this to avoid confusion, thoughts?

crucible-install.sh
```diff
@@ -0,0 +1,171 @@
...
+    if [ -z "$CRUCIBLE_EMAIL" ]; then
+        echo "Please enter your email address:"
+        read CRUCIBLE_EMAIL
+        echo "CRUCIBLE_EMAIL=\"$CRUCIBLE_EMAIL\"" >>"$identity"
+    fi
+}
+
+
+function has_dependency {
+    has_dep=0
```
k-rister, line 68: Indentation seems messed up in this function...

crucible-install.sh
```diff
@@ -0,0 +1,171 @@
...
+    exit_error "Failed to execute crucbile-project install, check $GIT_INSTALL_LOG for details" $EC_FAIL_INSTALL
+
+if [ ! -f $CRUCIBLE_AUTH_FILE ]; then
+    exit_error "Crucible authentication file not found. See --auth-file for details." $EC_FAIL_AUTH
+fi
+
+# native crucible install script already created this, only append
+cat << _SYSCFG_ >> $SYSCONFIG
+CRUCIBLE_USE_CONTAINERS=1
+CRUCIBLE_USE_LOGGER=0
```
k-rister, line 162: I would like to move to the loger being enabled by default.

k-rister (changes requested):

crucible-install.sh
rafaelfolco: oops

crucible-install.sh
rafaelfolco: ack

rafaelfolco:
> I wonder if we shouldn't move/rename bin/install as part of this to avoid confusion, thoughts?
Maybe some brief description in the README? I don't think we need to rename it.

crucible-install.sh
```diff
@@ -0,0 +1,173 @@
...
+# vim: autoindent tabstop=4 shiftwidth=4 expandtab softtabstop=4 filetype=bash
+# -*- mode: sh; indent-tabs-mode: nil; sh-basic-offset: 4 -*-
+
+# Installer Settings
+SYSCONFIG="/etc/sysconfig/crucible"
+DEPENDENCIES="podman git"
+INSTALL_PATH="/opt/crucible"
+GIT_REPO="https://github.com/perftool-incubator/crucible.git"
+GIT_INSTALL_LOG="/tmp/crucible-git-install.log"
+CRUCIBLE_AUTH_FILE="/root/.docker/crucible-client-server.json"
```
k-rister, line 11: I wonder if we should avoid setting a default like this.  In the proposed INSTALL.md it doesn't even mention the --auth-file option and then with this default it almost seems like it is setup to assume/hope that the user already has this setup and ready to go.  In reality, that seems highly unlikely since it is going to be specific to Crucible -- if Crucible is not installed why would they already have this?

k-rister (changes requested):

crucible-install.sh
rafaelfolco: agreed

k-rister (approved):
LGTM

crucible-install.sh
atheurer: agreed

atheurer:
> I wonder if we shouldn't move/rename bin/install as part of this to avoid confusion, thoughts?

Should it be renamed something like subprojects-install?

crucible-install.sh
atheurer: You are 100% certain logger if working fine?  If so, then I guess I am OK with that.

k-rister: I guarantee that it works ;-)

k-rister:
I like the sound of subprojects-install, plus the name is probably more appropriate anyway.

rafaelfolco:
> > I wonder if we shouldn't move/rename bin/install as part of this to avoid confusion, thoughts?
> 
> Should it be renamed something like subprojects-install?
Good idea.

k-rister:
I think it looks good and is ready to merge, agreed?

rafaelfolco:
> I think it looks good and is ready to merge, agreed?
From my side, let me manually test this latest change and make sure it works. We should include some automated tests that run locally inside a container so anyone is able to verify the change.

rafaelfolco:
> > I think it looks good and is ready to merge, agreed?
> > From my side, let me manually test this latest change and make sure it works. We should include some automated tests that run locally inside a container so anyone is able to verify the change.

For now I what can tell is "it works on my machine" :)
I'll be proposing an automated test that we can run and verify the change. Thanks for the reviews.

PR #117, opened 2021-03-12
Gitlab ci

rafaelfolco:
Initial ci includes basic functional tests for the crucible installer.


.gitlab-ci.yml
```diff
@@ -0,0 +1,41 @@
+test-installer:
+  stage: test
+  image: centos:8
```
k-rister, line 3: Can we run this against Centos Streams instead?  I think Andrew has the necessary bits to convert a Centos 8 to Streams.  Seems wise since that is the future...

.gitlab-ci.yml
```diff
@@ -0,0 +1,41 @@
+test-installer:
```
k-rister, line 1: What causes this particular test to be run as opposed to some other test (for example a future test would cause a Crucible update to run)?

.gitlab-ci.yml
```diff
@@ -0,0 +1,41 @@
...
+  stage: test
+  image: centos:8
+  script:
+    - |
+      set +e
+      set -x
+
+      # usage
+      stdout=$(bash crucible-install.sh --help)
```
k-rister, line 10: Small nit...do we have to run it as `bash crucible-install.sh` as opposed to `./crucible-install.sh`?  I suspect most users would use the `./crucible-install.sh` form so that seems like what we should test.

k-rister (commented):
I'm curious how other tests would look and what actually causes these tests to be run.

.gitlab-ci.yml
rafaelfolco: > I'm curious how other tests would look and what actually causes these tests to be run.
Fair point. I'll include at least a placeholder in the next commit.

.gitlab-ci.yml
rafaelfolco: Gitlab has something called 'stages'. 
stages:
 - installer
 - update
 - performance
In this example, performance stage will execute only if installer and update stages succeeded.

You can also specify tests that run if a particular file is changed:
```
only: 
    refs:
      - master
    changes: 
      - crucible-installer.sh
      - bin/crucible
```

.gitlab-ci.yml
rafaelfolco: Agreed. chmod +x on it now :)

.gitlab-ci.yml
rafaelfolco: > Can we run this against Centos Streams instead? I think Andrew has the necessary bits to convert a Centos 8 to Streams. Seems wise since that is the future...
I believe 'centos-stream' is the image, need to test.

rafaelfolco:
@k-rister @atheurer this is the poc for ci, please review. This is how the pipeline would look like https://gitlab.com/rafaelfolco/crucible/-/pipelines/271456533.

k-rister (commented):
I think this all looks good for a starting point at getting a CI working, but I do have one question.  Do we have to place the ".gitlab*" files in the project root or can we bury them somewhere (tests?)?  Since our actual use case for deploying Crucible is to clone the git repo I think it would make sense to keep the project root as clean as possible to avoid user confusion (and I need to look and see if this means anything else should move...).

rafaelfolco:
> I think this all looks good for a starting point at getting a CI working, but I do have one question. Do we have to place the ".gitlab*" files in the project root or can we bury them somewhere (tests?)? Since our actual use case for deploying Crucible is to clone the git repo I think it would make sense to keep the project root as clean as possible to avoid user confusion (and I need to look and see if this means anything else should move...).

Yes, we can place gitlab file anywhere, and it can be called anything: https://docs.gitlab.com/ee/ci/pipelines/settings.html#custom-cicd-configuration-path
/tests/.gitlab-ci.yml for general config... /tests/.test-installer.yml

k-rister:
> > I think this all looks good for a starting point at getting a CI working, but I do have one question. Do we have to place the ".gitlab*" files in the project root or can we bury them somewhere (tests?)? Since our actual use case for deploying Crucible is to clone the git repo I think it would make sense to keep the project root as clean as possible to avoid user confusion (and I need to look and see if this means anything else should move...).
> 
> Yes, we can place gitlab file anywhere, and it can be called anything: https://docs.gitlab.com/ee/ci/pipelines/settings.html#custom-cicd-configuration-path
> /tests/.gitlab-ci.yml for general config... /tests/.test-installer.yml

Ok, lets do that then.

k-rister (approved):

atheurer (approved):

rafaelfolco:
moved to https://gitlab.com/perftool-incubator/crucible
fixing badge...

rafaelfolco:
@k-rister this pr now is ready to merge. I moved the ci project to https://gitlab.com/perftool-incubator/crucible and updated the badge in README.md.

PR #123, opened 2021-03-25
Switch to shell executor

rafaelfolco:
A specific runner has been registered with the shell executor
to run ci jobs against crucible PRs.
This change removes the container image from the gitlab config
that was used by docker executor in shared runners.


k-rister:
So this should be merged, right?

rafaelfolco:
> So this should be merged, right?
ship it !

PR #127, opened 2021-04-05
Run crucible-install as root

rafaelfolco:
Running as non-root will fail. Adding a test to cover this
scenario.


rafaelfolco:
this has been included/merged into https://github.com/perftool-incubator/crucible/pull/128

PR #128, opened 2021-04-05
Make crucible config distro-generic

rafaelfolco:
Move from /etc/sysconfig/crucible to /etc/crucible.conf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # crucible
 
-## Pipeline Status
-
-[![pipeline status](https://gitlab.com/perftool-incubator/crucible/badges/master/pipeline.svg)](https://gitlab.com/perftool-incubator/crucible/-/commits/master)
+[![CI Actions Status](https://github.com/perftool-incubator/crucible/workflows/CI/badge.svg)](https://github.com/perftool-incubator/crucible/actions)
 
 ## Introduction
 
--- a/bin/_crucible_completions
+++ b/bin/_crucible_completions
@@ -4,8 +4,8 @@
 
 _crucible_completions() {
     if [ -z "$CRUCIBLE_HOME" ]; then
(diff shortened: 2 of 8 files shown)


atheurer:
How does this affect our internal installer?  Does the internal installer now use the public crucible-install.sh, so there are no updates to the internal installer for manipulating the config file?

rafaelfolco:
> How does this affect our internal installer? Does the internal installer now use the public crucible-install.sh, so there are no updates to the internal installer for manipulating the config file?
This is hardcoded in multiple places. I've changed several files in this commit: https://github.com/perftool-incubator/crucible/pull/128/commits/2cee04780e3860dbc433c64afd8a77f4fbe00b53. If the internal installer shared the same vars, we could change in the external installer only.

atheurer:
> If the internal installer shared the same vars, we could change in the external installer only.

I was thinking that the internal installer could just call the external installer with the login credentials provided, so the internal installer does not have to know any of these details about where the config file is.

Also, what happens to existing installations that still have /etc/sysconfig/crucible?  Is there a check to find this and move it to /etc/crucible.conf?

rafaelfolco:
> I was thinking that the internal installer could just call the external installer with the login credentials provided, so the internal installer does not have to know any of these details about where the config file is.

This looks odd, but I suppose I am missing something here. The external must call the internal, not the opposite ?

> Also, what happens to existing installations that still have /etc/sysconfig/crucible? Is there a check to find this and move it to /etc/crucible.conf?

If you installed crucible config on /etc/sysconfig, your crucible on /opt/crucible will still use it. If you run a new install, the new config on /etc/crucible.conf will be used. Old installer, old config, New installer, new config. If you want to use an old installation with the new config, you could manually copy it.

PR #131, opened 2021-04-06
Improve the efficiency and correctness of the update process

k-rister:
- The update process will now update Crucible twice -- the first time
  all changes are pulled down and the second time any new changes to
  the update process are ensured to run.

- By reordering the code only the Crucible updating happens twice, the
  subprojects are not subject to the same potential problems that
  Crucible is so they do not need special handling.  Running the
  updates on the subprojects multiple times would be at best
  inefficient and at worst may cause problems (ie. throttling) due to
  excessive git traffic to the respository host.

- Fix the help message to be more comprehensive and correctly
  formatted.

### DIFF
--- a/bin/update
+++ b/bin/update
@@ -18,21 +18,25 @@ fi
 
 . $CRUCIBLE_HOME/bin/base
 
+help_column_width=25
+
 function help() {
     echo "Usage:"
-    echo "  update <repository>"
-    echo ""
-    echo "The following repositories are supported"
-    echo ""
-    echo "all             |  Update all repositories (default)"
-    echo "crucible        |  Update the main crucible repository"
+    echo "  update <target>"
+    echo
+    echo "The following targets are supported:"
+    echo
+    printf "%-${help_column_width}s | Update all repositories (default)\n" "all"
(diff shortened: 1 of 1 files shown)


bin/update
```diff
@@ -44,28 +48,54 @@ else
...
+
+# beyond this point the script behaves "normally", meaning that it
+# does not have to create a copy of itself anymore and exec that copy
 
 case "${repo}" in
-    "all"|"crucible"|"controller-image")
-        # nothing to do here yet
-        foo=1
+    "____all"|"____crucible")
+        # any crucible specific post update logic goes here
```
rafaelfolco, line 78: this could be : as well

rafaelfolco (approved):
lgtm - I think I understand the workflow here, I am looking for how to test it and add a quick section to readme on how to update crucible and its sub-projects.

bin/update
k-rister: Not sure what you mean about the syntax.  Logic wise, since we know we want to do some post update file migration very soon I went ahead and noted where that should be done.

rafaelfolco: Sorry, I just meant using the built-in no-op ':' to "do nothing". Something like:
    "____all"|"____crucible")
        : echo noop

PR #140, opened 2021-04-29
Fix identity

rafaelfolco:
Append identity variables to file when args are provided.
Include a test to cover this scenario.

### DIFF
--- a/tests/test-installer
+++ b/tests/test-installer
@@ -52,4 +52,5 @@ sudo ./crucible-install.sh \
   --auth-file /tmp/auth-file.json \
   --name "Nobody" \
   --email "nobody@somewhere.com"
-sudo grep "CRUCIBLE_NAME" $idfile || exit 1
+sudo grep "CRUCIBLE_NAME=\"Nobody\"" $idfile || exit 1
+sudo grep "CRUCIBLE_EMAIL=\"nobody@somewhere.com\"" $idfile || exit 1
--- a/tests/test-installer
+++ b/tests/test-installer
@@ -40,3 +40,16 @@ test "$?" = "0" || exit 1
 grep "myregistry.io" $cfgfile || exit 1
 grep "auth-file.json" $cfgfile || exit 1
 command -v crucible || exit 1
+
+# identity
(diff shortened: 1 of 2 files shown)


k-rister:
Why is the CI check failing?

rafaelfolco:
> Why is the CI check failing?

fixed permissions for the identity file

tests/test-installer
```diff
@@ -40,3 +40,16 @@ test "$?" = "0" || exit 1
...
+touch /tmp/auth-file.json || true
+cfgfile=$(grep SYSCONFIG= crucible-install.sh | cut -d '=' -f2 | sed 's/"//g')
+idfile=$(grep IDENTITY= crucible-install.sh | cut -d '=' -f2 | sed 's/"//g')
+# TODO(rfolco): temporary workaround until we make it distro generic
+sudo mkdir -p $(dirname $cfgfile)
+sudo ./crucible-install.sh \
+  --registry myregistry.io/crucible \
+  --auth-file /tmp/auth-file.json \
+  --name "Nobody" \
+  --email "nobody@somewhere.com"
```
k-rister, line 54: Unless I'm not understanding things I don't think this worked in the CI test.  It says the test passed, but I don't see these values in the following grep output:

https://github.com/perftool-incubator/crucible/pull/140/checks#step:3:128

rafaelfolco:
@k-rister  I had to change the logic around identity a little. Please see if this is reasonable.

k-rister (approved):

k-rister:
Can you briefly describe what you had to change?  Seems like it would good understand that going forward...

rafaelfolco:
> Can you briefly describe what you had to change? Seems like it would good understand that going forward...

Summary:
Logic around identity has become a little complex because we may end up with many scenarios there. 

The problem:
Identity was overriding args specified from cli and the test was failing.

Fix:
I changed the precedence of loading name and email config. If --name or --email are specified, then skip loading identity file. Then check if name and email are specified (no matter if it was from args or from file) and read as input otherwise.
At the end, override changes in the identity file.

PR #146, opened 2021-06-28
Add command to completely reindex all runs

atheurer:
-add some missing help output


bin/_help
```diff
@@ -16,6 +16,9 @@ function help() {
     echo "console         |  Run user supplied programs inside a crucible wrapper for logging purposes"
     echo "wrapper         |  Run a crucible subproject command directly, within a crucible container"
     echo "get             |  Get either a result summary (get result) or any metric (get metric)"
+    echo "index           |  Index a result which is already post-processed with run-result-dir argument immediately following 'index'"
+    echo "postprocess     |  Post-process (and index) a result with run-result-dir argument immediately following 'index'"
+    echo "reindex         |  Completely delete and init ES, then post-process and index all runs found in $var_crucible/run"
```
k-rister, line 21: Should this change slightly to `index [all]` and `postprocess [all]`?

Given the fact that we currently have `index` and `postprocess` where `postprocess` does both, it seems odd that `reindex` would also do both while `index` does not.  If it was switched to both arguments having an optional `all` parameter then the behavior between operating on one or many results could be the same.

I'm just thinking that consistency will make it easier to use...

bin/_main
```diff
@@ -232,6 +232,21 @@ elif [ "${1}" == "postprocess" ]; then
     shift
     base_run_dir=$(echo "$@" | sed -e "s/--base-run-dir\(\s\+\|=\)//")
     post_process_run "${base_run_dir}"
+elif [ "${1}" == "reindex" ]; then
+    shift
+    start_es
+    delete_es_cmd="${CRUCIBLE_HOME}/subprojects/core/CommonDataModel/templates/delete.sh"
+    init_es_cmd="${CRUCIBLE_HOME}/subprojects/core/CommonDataModel/templates/init.sh"
+    $podman_run --name crucible-es-delete "${container_common_args[@]}" "${container_rs_args[@]}" $CRUCIBLE_CONTAINER_IMAGE $delete_es_cmd
+    $podman_run --name crucible-es-init "${container_common_args[@]}" "${container_rs_args[@]}" $CRUCIBLE_CONTAINER_IMAGE $init_es_cmd
```
k-rister, line 241: We should probably get in the habit of including unique identifiers in pod names so that we don't run into an issue where they conflict if invoked multiple times...of course in this particular case since you are doing a global action maybe it should conflict -- but if that is the case then we should probably check a return code and bail out if it fails to launch.

bin/_help
atheurer: Yeah, I figured we would refine this a bit, but I wanted something in there for now in case others need to reindex everything.

bin/_main
atheurer: Agreed

PR #152, opened 2021-07-09
Fix install and ci tests

rafaelfolco:
check missing auth before running subprojects/install
Add test to override existing installation

### DIFF
--- a/crucible-install.sh
+++ b/crucible-install.sh
@@ -152,6 +152,10 @@ for dep in $DEPENDENCIES; do
     has_dependency $dep
 done
 
+if [ ! -f $CRUCIBLE_AUTH_FILE ]; then
+    exit_error "Crucible authentication file not found. See --auth-file for details." $EC_AUTH_FILE_NOT_FOUND
+fi
+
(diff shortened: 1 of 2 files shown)


crucible-install.sh
```diff
@@ -152,6 +152,10 @@ for dep in $DEPENDENCIES; do
     has_dependency $dep
 done
 
+if [ ! -f $CRUCIBLE_AUTH_FILE ]; then
+    exit_error "Crucible authentication file not found. See --auth-file for details." $EC_AUTH_FILE_NOT_FOUND
+fi
+
 if [ -d $INSTALL_PATH ]; then
     old_install_path="/opt/crucible-moved-on-`date +%d-%m-%Y_%H:%M:%S`"
```
k-rister, line 160: `date +%Y-%m-%d_%H:%M:%S`?

k-rister (approved):

crucible-install.sh
rafaelfolco: > `date +%Y-%m-%d_%H:%M:%S`?

Is this a feature request ? :)

PR #163, opened 2021-08-05
ensure unique names for pods that could exist multiple times simultan…

k-rister:
…eously


rafaelfolco (approved):
superficial check, I am not very familiar with pods creation

PR #165, opened 2021-08-10
only run ci jobs on pull requests, not pushes

k-rister:
- this avoids unneccessary runs of the ci


rafaelfolco (approved):
we can restrict events as well if needed...
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

PR #176, opened 2021-09-13
Dev kmr

k-rister:


k-rister:
I would appreciate as many eyes as possible on this PR due to the fact that is has the potential to break existing installations.

bin/subprojects-install
```diff
@@ -72,6 +72,35 @@ done
...
+    active=0
+    if [ -n "${subprojects[$name_link]}" ]; then
+        sp_type=`echo ${subprojects[$name_link]} | awk '{print $2}'`
+        case "$sp_type" in
+            "benchmark")
+                if [ "${type_dir}" == "benchmarks" ]; then
+                    active=1
+                fi
+                ;;
+            "tool")
```
rafaelfolco, line 87: wondering why "benchmark" and "tool" without 's' here and in the directory it is with 's'
```
subprojects/benchmarks/hwlatdetect
subprojects/benchmarks/cyclictest
subprojects/benchmarks/uperf
subprojects/benchmarks/tracer
subprojects/benchmarks/trafficgen
subprojects/benchmarks/oslat
subprojects/benchmarks/fio
subprojects/tools/ovs
subprojects/tools/tcpdump
subprojects/tools/ftrace
subprojects/tools/procstat
subprojects/tools/kernel
subprojects/tools/sysstat
subprojects/core/multiplex
subprojects/core/packrat
subprojects/core/workshop
subprojects/core/junk-drawer
subprojects/core/toolbox
subprojects/core/roadblock
subprojects/core/CommonDataModel
subprojects/core/rickshaw
```

rafaelfolco (approved):
lgtm, just make sure the missing 's' is ok

bin/subprojects-install
k-rister: I agree that it does seem "odd", but its actually an inversion of the existing logic:

https://github.com/perftool-incubator/crucible/blob/master/bin/subprojects-install#L95-L105

k-rister: which I guess are influenced with the format that was chosen for this file originally:

https://github.com/perftool-incubator/crucible/blob/master/config/default_subprojects

PR #178, opened 2021-09-21
try the ci mechanism for testing crucible and rickshaw updates

tc-wilson:


k-rister:
@tc-wilson is this PR still valid?

PR #180, opened 2021-09-28
add support for setting --tls-verify=false for container registry ope…

k-rister:
…rations

- this allows for insecure registries to be used, although there are
  other details that may be required such as endpoint support

### DIFF
--- a/crucible-install.sh
+++ b/crucible-install.sh
@@ -23,13 +23,18 @@ EC_FAIL_INSTALL=5
 EC_AUTH_FILE_NOT_FOUND=6
 EC_FAIL_DEPENDENCY=7
 EC_FAIL_REGISTRY_UNSET=8
-#9
+EC_TLS_VERIFY_ERROR=9
 EC_INVALID_OPTION=10
 EC_UNEXPECTED_ARG=11
 EC_FAIL_REGISTRY_SET=12
 EC_FAIL_AUTH_SET=13
 EC_FAIL_CHECKOUT=14
 EC_PUSHD_FAIL=15
+
+# remove a previous installation log
+if [ -e ${GIT_INSTALL_LOG} ]; then
+    rm ${GIT_INSTALL_LOG}
+fi
 
 function determine_git_install_source {
     local PWD SCRIPT_DIR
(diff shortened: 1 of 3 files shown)


tests/test-installer
```diff
@@ -22,6 +22,13 @@ sudo ./crucible-install.sh \
   --client-server-auth-file /tmp/auth-file.json
 test "$?" = "$ec" || exit 1
 
+# invalid tls verify value
+ec=$(grep EC_TLS_VERIFY_ERROR= crucible-install.sh | cut -d '=' -f2)
+sudo ./crucible-install.sh \
+  --client-server-registry myregistry.io/crucible \
+  --client-server-tls-verify foo
```
rafaelfolco, line 29: negative test, that's good :)

crucible-install.sh
```diff
@@ -23,14 +23,19 @@ EC_FAIL_INSTALL=5
...
 EC_FAIL_REGISTRY_SET=12
 EC_FAIL_AUTH_SET=13
 EC_FAIL_CHECKOUT=14
 EC_PUSHD_FAIL=15
 
+# remove a previous installation log
+if [ -e ${GIT_INSTALL_LOG} ]; then
+    rm ${GIT_INSTALL_LOG}
+fi
+
```
rafaelfolco, line 38: why? what does it have to do with tls ?

crucible-install.sh
```diff
@@ -142,6 +147,7 @@ function usage {
 
     optional:
         --client-server-auth-file <authentication file>
+        --client-server-tls-verify true|false
```
rafaelfolco, line 150: --client-server-tls-verify  could be a positional arg

tests/test-installer
rafaelfolco: missing a successful case at least

rafaelfolco (approved):
lgtm, minor things...

crucible-install.sh
k-rister: Nothing.  Except I was initially getting a totally misleading error log because it was discovering an old one :(

PR #181, opened 2021-10-04
add engine subproject

k-rister:


rafaelfolco (approved):
why do we need a separate component for this ?
https://github.com/perftool-incubator/engine/blob/main/README.md

k-rister:
I'm preparing to move the engine pieces from the rickshaw project into their own dedicated subproject which is a long standing TODO

PR #186, opened 2021-10-13
use bin/result-processor.py to provide tab completion data

k-rister:


bin/_main
```diff
@@ -278,10 +278,19 @@ if [ "${1}" == "log" ]; then
     shift
     crucible_log ${1} ${LOG_DB}
     EXIT_VAL=$?
-elif [ "${1}" == "ls" -o "${1}" == "tags" ]; then
+elif [ "${1}" == "ls" -o "${1}" == "tags" -o "${1}" == "result-completion" ]; then
```
rafaelfolco, line 281: result-completion ? isn't just "completion" ?

bin/result-processor.py
```diff
@@ -47,6 +47,17 @@ def process_options():
                                        required = True)
 
 
+    parser_completion = subparsers.add_parser("completion",
```
rafaelfolco, line 50: unless I am not following, it is "completion"

bin/_main
k-rister: It comes down to a context issue.  I have wondered if I chose the right way, but for the moment I have decided to stay with it.  I'm curious if you agree when I explain my reasoning.

In the scope of crucible, it is helpful to denote the crucible command (`crucible result-completion`) as specific to results in case we expand our tab completion querying in the future for things other than results (using some other script(s)).  Image something like `crucible benchmark-completion` as a future improvement (if you look at the `_crucible_completions` file you will see we are doing that with some find magic today).

For the script that generates the tab completion data (in this case `result-processor.py`) it somewhat redundant to label it `result-completion` when all it handles is results.

So ultimately, when you ask crucible for help with result-completion it knows to ask the result processor script for its completion output.

Thoughts?

rafaelfolco: Thanks for explaining. Sounds reasonable.

rafaelfolco (approved):

PR #191, opened 2021-11-12
detect created but not running containers and clean them up

k-rister:
- Sometimes podman gets in a state where containers that should be
  removed when they stop are left around, this is usually tied to some
  undesirable event (such as a system crash), and when this happens
  our logic to start the containers if they are not present fails.

- If a container is detected in this state attempt to forcibly remove
  it so that it can be relaunched through the normal container start
  procedures.


mrbojangles3:
LGTM

atheurer (approved):

PR #197, opened 2022-02-09
improve tool configuration via the run command

k-rister:


atheurer (approved):

bin/_main
```diff
@@ -508,20 +512,27 @@ elif [ "${1}" == "run" ]; then
...
+            echo "\"--tool-params=<file>\"."
+            echo ""
+            echo "You can also disable tools entirely using the \"--no-tools\" parameter."
+        else
+            cp ${tool_params} ${tool_params_file}
+        fi
     else
-        cp ${tool_params} ${tool_params_file}
+        echo "Disabling all tools due to --no-tools"
+        echo "[]" > ${tool_params_file}
```
HughNhan, line 535: So .. we will have this behavior ... there is a "custom" tools-params.json in-place, but the user just want to run --no-tools for this one run for whatever reason. Once the --no-tools run runs, the original tools-params.json will be overwritten with "[]".  Is that right?

k-rister: Yes.  The tool-params.json that is created via the run command (either via copying the default set or by this new method) is always per run.  There is no persistence from run-to-run -- besides whatever files the user themselves has created in their directory/directories.

atheurer: So your original tool params file is safe/unmodified

HughNhan (commented):
LGTM. Thanks

PR #208, opened 2022-04-01
add support for processing multiple arguments with archive and unarchive

k-rister:


rafaelfolco (approved):
moving archive to the end seems a good solution

PR #209, opened 2022-05-09
update firewall port opening logic to be more flexible

k-rister:
- the "public" zone is not always the default zone that needs to be
  configured (ie. Fedora 35), so automatically detect the default
  firewall zone and use it when opening ports

- create some very basic infrastructure to allow for more flexible
  firewall configuration in the future (ie. environments that do not
  use firewall-cmd/firewalld)

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -41,12 +41,31 @@ container_rs_args+=("-e RS_NAME=\"$CRUCIBLE_NAME\"")
 container_rs_run_args=()
 container_rs_run_args+=("--mount=type=bind,source=/var/lib/containers,destination=/var/lib/containers")
 
+function open_firewall_port() {
+    local port
+    local protocol
+    local firewall_mgmt
+
+    port=${1}; shift
+    protocol=${1}; shift
+    firewall_mgmt="firewall-cmd"
+
+    case "${firewall_mgmt}" in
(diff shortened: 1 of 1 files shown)


mrbojangles3:
Would we have to check if the interface we want is inside the default zone?

k-rister:
> Would we have to check if the interface we want is inside the default zone?

There is a significant amount of assumptions that are being made in this code right now, most prominently that opening the requested port in the default zone is all that we need.  Behavior wise, this PR is only adding the ability to detect what the default zone is rather than assuming it is named default.

I did attempt to create a new zone named Crucible and only add the ports that we wanted open to that zone.  I was also going to add all interfaces known to firewalld to that zone, however firewalld would only allow an interface to exist in a single zone so I couldn't add the interface that was already tied to the default zone.

Ultimately, there are limits to what we can do here automatically.  If we want to allow for more complex setups such as only opening the ports on a certain interface we are going to have to add a lot more logic and the ability to pass configuration information to these functions.  None of that exists right now.

In particular, this PR is really just trying to address the fact that my Fedora 35 based controller wouldn't work after a reboot given the default firewall configuration.  This PR fixes that with minimal changes overall.

mrbojangles3 (approved):

PR #225, opened 2022-07-27
change podman's log level to suppress warnings

k-rister:
- This doesn't appear to suppress all warnings so the grep filter is
  still necessary, I assume this is because a library that podman is
  loading is emitting warnings that don't honor the log level.


k-rister:
#227

atheurer (approved):

PR #231, opened 2022-08-02
Dev kmr

k-rister:


k-rister:
Seems like I still need to fix something here with the hard reset code.

bin/base
```diff
@@ -751,3 +751,27 @@ function init_es() {
 
     return ${RC}
 }
+
+function confirm_continue() {
+    local ref
+    local action_message
+    local selection
+
+    typeset -n ref=${1}
```
rafaelfolco, line 774: -n is bad option for me, are you sure this works ?

rafaelfolco (approved):

bin/base
k-rister: it does, assuming you have a new enough bash.  we are actually forcefully installing a newer version of bash into the client-server containers in order to support this syntax already (since our oldest usernevs don't support it natively)

k-rister: Additionally, in this situation a prompt is being printed and then the answer is being captured so we can't use traditional bash output capture since that would include the prompt.

PR #237, opened 2022-08-26
exec workshop in build-controller.sh

k-rister:
- pass control over to workshop entirely

- ensure return codes are properly given to the call

### DIFF
--- a/workshop/build-controller.sh
+++ b/workshop/build-controller.sh
@@ -15,9 +15,7 @@ fi
 if pushd ${CRUCIBLE_HOME} > /dev/null; then
     export TOOLBOX_HOME=${CRUCIBLE_HOME}/subprojects/core/toolbox
     
-    ./subprojects/core/workshop/workshop.pl --userenv ./workshop/fedora36.json --requirements ./workshop/controller-workshop.json --label crucible-controller
-
-    popd > /dev/null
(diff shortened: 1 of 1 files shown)


workshop/build-controller.sh
```diff
@@ -15,9 +15,7 @@ fi
 if pushd ${CRUCIBLE_HOME} > /dev/null; then
     export TOOLBOX_HOME=${CRUCIBLE_HOME}/subprojects/core/toolbox
     
-    ./subprojects/core/workshop/workshop.pl --userenv ./workshop/fedora36.json --requirements ./workshop/controller-workshop.json --label crucible-controller
-
-    popd > /dev/null
+    exec ./subprojects/core/workshop/workshop.pl --userenv ./workshop/fedora36.json --requirements ./workshop/controller-workshop.json --label crucible-controller
```
rafaelfolco, line 18: exec will just call the command in a separate shell, isn't it ? not sure what was the issue

rafaelfolco (approved):

workshop/build-controller.sh
k-rister: `exec` "replaces" the currently running process with the new one without doing a fork.  In this context, the way that the script was originally written the return code from `workshop.pl` was being ignored and not returned to the user/caller.  Since this script is really just a wrapper to ensure the proper environment is in place before launching `workshop.pl` I figured that I can use `exec` to launch `workshop.pl` and then I don't have to deal with passing return codes back or any other complications with the wrapper any more.

PR #239, opened 2022-08-30
build controller ci testing

k-rister:


.github/workflows/crucible-ci.yaml
```diff
@@ -13,7 +13,35 @@ on:
...
+        diff_cmd="git diff --name-only HEAD^1 HEAD"
+        echo "Files changed:"
+        ${diff_cmd}
+        echo
+        workshop_files_changed=$(${diff_cmd} | grep "^workshop/" | wc -l)
+        echo "workshop_files_changed=${workshop_files_changed}"
+        echo
+        if [ ${workshop_files_changed} -gt 0 ]; then answer="yes"; else answer="no"; fi
+        echo "Setting answer to '${answer}'"
+        echo "::set-output name=answer::${answer}"
```
rafaelfolco, line 40: have you seen this ? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

This seems to do the same as you are doing manually...

rafaelfolco (commented):
paths: may save you some logic there

.github/workflows/crucible-ci.yaml
k-rister: I'm already using that in this workflow.  What I was trying to do here was avoid having to create yet another workflow that only had a single parameter difference from this one.  By adding this job and creating an output from it that it used as input to the other jobs I was able to keep a single workflow with all the other logic consolidated.

rafaelfolco: oh I see, any reason you want one single workflow ?

k-rister: To avoid replication and having to keep the workflows synchronized.

rafaelfolco: simplicity, fair enough

rafaelfolco (approved):

PR #240, opened 2022-08-30
Initial support for single json config

rafaelfolco:
crucible run --from-file <config.json>

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -219,6 +219,15 @@ elif [ "${1}" == "run" ]; then
             "--no-tools")
                 no_tools=1
                 ;;
+            "--from-file")
+                run_file=${1}
+                mv_params=${base_run_dir}/mv_params.json
+                tool_params=${base_run_dir}/tool-params.json
+                python3 ${TOOLBOX_HOME}/bin/get-json-config.py --json ${run_file} --mv-params > $mv_params
+                python3 ${TOOLBOX_HOME}/bin/get-json-config.py --json ${run_file} --tool-params > $tool_params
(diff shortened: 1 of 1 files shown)


bin/crucible-run.json.example
```diff
@@ -0,0 +1,31 @@
...
+                    { "arg": "bs", "vals": [ "4K", "16K" ] }
+                ]
+            }
+        ]
+    },
+    "tool-params": {
+    },
+    "passthru-args": ""
+}
+
```
k-rister, line 31: The `bin` directory seems like an odd location for this file.

bin/get_config.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/python3
```
k-rister, line 1: I'm thinking that this belongs in https://github.com/perftool-incubator/toolbox/tree/main/bin, right?  Eventually we are going want to be able to extract things from this configuration file in a bunch of different places and putting a tool like this in toolbox is the best way to ensure it's available.

We should also maybe use a slighly more descriptive name, `get_config.py` is pretty vague.

bin/crucible-run.json.example
rafaelfolco: ack

bin/get_config.py
rafaelfolco: agreed

k-rister (approved):

k-rister:
We should start thinking about creating test cases for this code in [`crucible-ci`](https://github.com/perftool-incubator/crucible-ci).

rafaelfolco:
> We should start thinking about creating test cases for this code in [`crucible-ci`](https://github.com/perftool-incubator/crucible-ci).

Indeed. This should not break anything, we're just adding a new option, not changing the current workflows.

PR #256, opened 2022-10-29
add concurrency limitations to the crucible-ci workflow

k-rister:
- create a concurrency group per pull request

- when a new execution of the workflow is queued for the same concurrency group as an existing execution, cancel the prior executions

- this avoids inefficient workflow execution via wasted runs


HughNhan (approved):
LGTM!

PR #269, opened 2022-12-05
Prevent passthru in cmdline when --run-file is used

rafaelfolco:
Do not allow user to specify passthru args in command line if --run-file <all-in-one.json> is used. Passthru args must be set in the json file instead.

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -246,9 +246,12 @@ elif [ "${1}" == "run" ]; then
         esac
     done
 
-    # merge passthru args from json file
+    if [ -n "${passthru_str}" && -n "${passthru_args}" ]; then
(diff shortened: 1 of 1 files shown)


k-rister (commented):
This is only going to catch pass through arguments that are supplied AFTER the `--from-file` option is specified, but not before...correct?

rafaelfolco:
> This is only going to catch pass through arguments that are supplied AFTER the `--from-file` option is specified, but not before...correct?

True, thanks for catching this. I know what I should do here then.

HughNhan (approved):
Looks like all concerned parties have spoken, and the CI went thru. Ship it!

k-rister (approved):

PR #276, opened 2022-12-09
switch to reusable crucible-ci core workflow

k-rister:


.github/workflows/crucible-ci.yaml
```diff
@@ -10,205 +10,12 @@ on:
...
-  crucible-ci-complete:
-    runs-on: ubuntu-latest
-    needs:
-    - crucible-ci-github-runners
-    - crucible-ci-self-hosted-runners
-    steps:
-    - name: Confirm Success
-      run: echo "crucible-ci-complete"
+  call-core-crucible-ci:
+    uses: perftool-incubator/crucible-ci/.github/workflows/core-crucible-ci.yaml@main
```
rafaelfolco, line 15: @main is really required ? can we use any branch ?

rafaelfolco (approved):

.github/workflows/crucible-ci.yaml
k-rister: Apparently there has to be *something* there...it can be a branch name, a version, or a specific commit hash that I know of.  When developing this stuff I was using the branch name that I was submitting my code to in the crucible-ci repo (`dev-kmr`) and once that code was merged up I removed the `@dev-kmr` thinking I didn't need it anymore but it complained that it required something so I put the `@main` there to always point to the HEAD of the crucible-ci repo.

PR #283, opened 2023-01-31
update the paths-ignore lists in the non faux workflows

k-rister:


rafaelfolco (approved):
these names are confusing

PR #294, opened 2023-02-28
Dev kmr

k-rister:


bin/_git-fetch
```diff
@@ -0,0 +1,79 @@
...
+    else
+        echo "ERROR: Could not pushd to ${CRUCIBLE_HOME}/subprojects"
+        exit 1
+    fi
+
+    CURRENT_TIMESTAMP=$(date +%s)
+    echo "${CURRENT_TIMESTAMP} ${UPDATE_COUNT}" > "${UPDATE_STATUS_FILE}"
+} > "${UPDATE_STATUS_FILE}.log" 2>&1
+
+rm "${LOCK_FILE}"
```
rafaelfolco, line 79: "rm -f" works in any env

bin/crucible
```diff
@@ -53,6 +53,17 @@ function setup_pipe() {
     mknod ${PIPE} p
 }
 
+if [ "${1}" != "update" ]; then
+    ${CRUCIBLE_HOME}/bin/_git-fetch &
+    if [ -e "${UPDATE_STATUS_FILE}" ]; then
+        read OTHER UPDATE_COUNT < "${UPDATE_STATUS_FILE}"
+        if [ "${UPDATE_COUNT}" != "0" ]; then
+            echo "*** NOTICE: Crucible is at least ${UPDATE_COUNT} commits behind.  See 'crucible repo info' for details. ***"
```
rafaelfolco, line 61: I like it

bin/crucible
```diff
@@ -53,6 +53,17 @@ function setup_pipe() {
     mknod ${PIPE} p
 }
 
+if [ "${1}" != "update" ]; then
+    ${CRUCIBLE_HOME}/bin/_git-fetch &
+    if [ -e "${UPDATE_STATUS_FILE}" ]; then
+        read OTHER UPDATE_COUNT < "${UPDATE_STATUS_FILE}"
```
rafaelfolco, line 59: what is this magic around OTHER ?

bin/_git-fetch
```diff
@@ -0,0 +1,79 @@
...
+            echo "REPO: ${tmp_repo}"
+            if pushd ${tmp_repo} > /dev/null; then
+                if [ -d "./.git" ]; then
+                    git remote --verbose update --prune
+
+                    git_status=$(git status --porcelain=2 --untracked-files=no --branch)
+                    git_tracking=$(echo "${git_status}" | grep "branch\.head" | awk '{ print $3 }')
+
+                    if [ -n "${git_tracking}" -a "${git_tracking}" != "(detached)" ]; then
+                        git_delta_behind=$(echo "${git_status}" | grep "branch\.ab" | awk '{ print $4 }' | sed -e "s/^-//" )
```
rafaelfolco, line 57: multiple greps and awk deserve a comment on what it is supposed to do, agreed ?

bin/_git-fetch
```diff
@@ -0,0 +1,79 @@
...
+fi
+
+touch "${LOCK_FILE}"
+
+if [ -e "${UPDATE_STATUS_FILE}" ]; then
+    read FETCH_TIMESTAMP OTHER < "${UPDATE_STATUS_FILE}"
+
+    CURRENT_TIMESTAMP=$(date +%s)
+
+    # re-run every 24 hours
```
rafaelfolco, line 34: comments!! beautiful!

bin/_git-fetch
```diff
@@ -0,0 +1,79 @@
...
+if [ ! -e ${CRUCIBLE_HOME} ]; then
+    echo "Could not find ${CRUCIBLE_HOME}, exiting."
+    exit 1
+fi
+
+source ${CRUCIBLE_HOME}/bin/base
+
+LOCK_FILE="${UPDATE_STATUS_FILE}.lock"
+
+if [ -e "${LOCK_FILE}" ]; then
```
rafaelfolco, line 23: just quit ?

k-rister: Yeah, this is just to prevent the updater from running multiple times simultaneously.  Since this is being spun off in the background it is possible for multiple commands to be run while the original backgrounded process is still running so this prevents additional copies from executing.  Having multiple copies executing isn't really a very bad thing, it's just very inefficient and sending unecessary traffic to the upstream repo.

bin/_git-fetch
k-rister: I suppose...

bin/_git-fetch
k-rister: Are you saying the that the `-f` may be necessary in some environments?

bin/crucible
k-rister: I debated this when I originally wrote it.  The file ends up having two values written to it, a timestamp and an update count.  In this instance I'm just reading the update count and don't care about the value in the timestamp field so I just made up an arbitrary variable name to stuff that timestamp value into.  I can make the variable name more accurate if you think it makes more sense that way.

bin/repo
```diff
@@ -74,12 +74,19 @@ function repo_info() {
 
         local current_commit=$(git rev-parse --short HEAD)
 
+        local git_delta_behind=$(echo "${git_status}" | grep "branch\.ab" | awk '{ print $4 }' | sed -e "s/^-//" )
```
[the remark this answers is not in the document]
k-rister: I reworked this to use common code that is documented.

bin/_git-fetch
k-rister: I reworked this to use common code that is documented.

atheurer (approved):

PR #307, opened 2023-09-06
Add multibench support to all-in-one json

rafaelfolco:
This patch integrates the multibench functionality into the crucible run files, aka all-in-one json. It includes single bench and multibench run mode when --from-file is used. Backwards compatibility takes care of the old run mode, which sets the benchmark from the cli with 'crucible run <benchmark(s)> [args]'.

A dry run mode is added through the env var CRUCIBLE_DRY_RUN=1 to help testing debugging param construction, parsing and validation.

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -156,12 +156,6 @@ elif [ "${1}" == "postprocess" ]; then
 elif [ "${1}" == "run" ]; then
     arg_list="$@"
     shift
-    benchmark=${1}
-    benchmarks=`echo ${benchmark} | sed -e 's/,/-and-/g'`
-    shift
-    base_run_dir="${var_run_crucible}/${benchmarks}--${datetime}--${SESSION_ID}"
-
-    mkdir -pv "$base_run_dir/config" >/dev/null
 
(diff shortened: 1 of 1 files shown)


bin/_main
```diff
@@ -170,57 +164,116 @@ elif [ "${1}" == "run" ]; then
...
+            exit_error "Error: blockbreaker rc=$? stdout=$tags_str"
+        fi
+
+        endpoint_str=$(python3 ${RICKSHAW_HOME}/util/blockbreaker.py --json ${run_file} --config endpoint)
+        if [ $? != 0 ]; then
+            exit_error "Error: blockbreaker rc=$? stdout=$endoint_str"
+        fi
+
+        passthru_blk=$(python3 ${RICKSHAW_HOME}/util/blockbreaker.py --json ${run_file} --config passthru
+-args)
```
k-rister, line 223: this seems weird...

rafaelfolco: fixing...

k-rister (dismissed):

k-rister (approved):

PR #310, opened 2023-09-28
convert blockbreaker.py calls from 'passthru-args' to 'run-params'

k-rister:


bin/_main
```diff
@@ -266,16 +266,16 @@ elif [ "${1}" == "run" ]; then
...
+        # extract run-params stream --key1 val1 --key2 val2 ...
+        blockbreaker_args="python3 ${RICKSHAW_HOME}/util/blockbreaker.py --json ${base_run_dir}/config/${run_file_json} --config run-params"
         blockbreaker_cmd="$podman_run -i --name crucible-blockbreaker-${SESSION_ID} "${container_common_args[@]}" $CRUCIBLE_CONTAINER_IMAGE ${blockbreaker_args}"
         blockbreaker_output=$(${blockbreaker_cmd})
         EXIT_VAL=$?
         if [ ${EXIT_VAL} != 0 ]; then
             exit_error "ERROR: blockbreaker failed to run with rc=$EXIT_VAL, stdout=${blockbreaker_output}"
         fi
-        passthru_str=$(sed -e 's/^"//' -e 's/"$//' <<< ${blockbreaker_output})
+        run_params_str=${blockbreaker_output}
```
rafaelfolco, line 278: is blockbreaker now returning the stream without double quotes ? Not sure this would be a problem, the stream will be appended to the crucible command like crucible run --from-file json "--key val --key val..."

k-rister: correct, no quotes...it is working fine in my testing

rafaelfolco (approved):
lgtm

PR #325, opened 2024-02-02
update file locking behavior for git status updating

k-rister:


k-rister:
@tc-wilson exposed some problems (when running multiple metric queries simultaneously) with the overly simplistic file presence based file locking that was originally used.  This patch makes use of flock to get more correct behavior.

atheurer (approved):

PR #346, opened 2024-06-23
add a workflow that will run on a schedule (or adhoc) to forceably re…

k-rister:
…build all engine images

### DIFF
--- a/.github/workflows/crucible-scheduled.yaml
+++ b/.github/workflows/crucible-scheduled.yaml
@@ -1,0 +1,17 @@
+name: crucible-scheduled
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/crucible-scheduled.yaml
```diff
@@ -0,0 +1,17 @@
+name: crucible-scheduled
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
```
rafaelfolco, line 6: workflow_dispatch allows manual trigger, and it will use the default settings I assume

k-rister: What do you mean by "default settings"?  If you manually trigger this workflow (through the web gui for example) it will run just like if it were scheduled to run at that time, the parameters to the workflow will be exactly the same as if it were running via the cron job like interface.

PR #348, opened 2024-06-27
[spec] Stable releases for Crucible installs

rafaelfolco:
Add a mechanism for installing stable releases that are "locked- down".

### DIFF
--- a/spec/stable-release.md
+++ b/spec/stable-release.md
@@ -1,0 +1,62 @@
+# Stable Release
+Implement a "Release" mechanism for installing "consistent" versions of 
+Crucible.
+
+## Problem description
+We need to be able to install a "version" of Crucible that is "locked down" 
+and won't change at all if it is installed over time.
+
+Crucible installations cannot be used for regression testing in a consistent
+manner.
+
+bin/update will override the installed repository by pulling the latest commit
+if the installed commit is an older one.
+
+Crucible installer clones Crucible repository and all sub-projects from the
+latest "main" branch. A default installation contains all the Crucible 
+repositories from the latest merged commits.
+
+```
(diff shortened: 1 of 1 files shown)


spec/stable-release.md
```diff
@@ -0,0 +1,62 @@
...
+#project-name        project-type   git-repo-url                     branch
+rickshaw             core           /rickshaw                        master
+multiplex            core           /multiplex                       master
+roadblock            core           /roadblock                       master
+...
+hwlatdetect          benchmark      /bench-hwlatdetect               main
+tracer               benchmark      /bench-tracer                    main
+...
+ftrace               tool           /tool-ftrace                     master
+testing              doc            /testing-repo                    master
```
k-rister, line 29: Do you think we should switch to the "new" subproject config idea (JSON) as part of this?

spec/stable-release.md
```diff
@@ -0,0 +1,62 @@
...
+...
+```
+
+## Proposed change
+"Locked down" release by pinning all sub-projects to a commit that is 
+verified by CI after code merges. Sub-projects includes core and benchmark
+repositories. Stable releases must be labeled to be an "out of the box"
```
k-rister, line 36: Not sure I follow what you mean by "out of the box" here...

spec/stable-release.md
rafaelfolco: I think so. Not a dependency, just a convenience, as we are changing crucible config/default_projects.

spec/stable-release.md
rafaelfolco: For regression testing, "out of the box" Crucible would be crucible-2024.2, not the upstream latest merged code. We can avoid this and just call it a stable release 2024.2.

PR #351, opened 2024-07-08
Release tag mechanism for locked-down installs

rafaelfolco:
Push a labeled version of crucible that passes ci after merging, including all pinned commits of the sub-projects. This release mechanism will be used for regression testing and also by users that need more determinsm from crucible stable installs that do not change over time.

Additional changes are required to implement the installation bits that will take advantage of this release tag mechanism.

### DIFF
--- a/.github/workflows/crucible-release.yaml
+++ b/.github/workflows/crucible-release.yaml
@@ -1,0 +1,203 @@
+name: crucible-release
+on:
+  schedule:
+    # Runs every 3 months (1st day of the quarter)
+    - cron: "0 0 1 */3 *"
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        required: false
+        type: boolean
+        description: Do NOT push/delete tag to/from the repositories
+        default: false
+    inputs:
+      custom-tag:
+        required: false
+        type: string
+        description: Custom tag to push/delete
+      action:
+        required: true
+        type: choice
+        options:
+        - push
+        - delete
+        description: Action to perform with the custom-tag (push or delete)
+        default: push
+
+jobs:
+  release-tag:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      tag: ${{ steps.gen-release-tag.outputs.tag }}
+    steps:
+    - name: Generate release tag
+      if: github.event_name != 'workflow_dispatch' || github.event.inputs.custom_tag == ''
+      id: gen-release-tag
+      run: |
+          year="$(date +%Y)"
+          month="$(date +%m)"
+          quarter="$(((month-1)/3+1))"
+          echo "tag=$year.$quarter" >> $GITHUB_OUTPUT
+    - name: Generate custom tag
+      if: github.event_name == 'workflow_dispatch' && github.event.inputs.custom_tag != ''
+      id: gen-custom-tag
+      env:
+          CUSTOM_TAG: ${{ inputs.custom-tag }}
+      run: |
+          echo "tag=$CUSTOM_TAG" >> $GITHUB_OUTPUT
+    - name: Get release tag
+      id: get-release-tag
+      env:
+          TAG: ${{ needs.release-tag.outputs.tag }}
+      run: echo "$TAG"
+
+  crucible-tag:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    - name: checkout crucible default
+      uses: actions/checkout@v4
+      with:
+        repository: perftool-incubator/crucible
+        ref: master
+        path: crucible
+    - name: get tags from the crucible repo
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin
+    - name: push if tag does not exist in the crucible repo
+      if: ${{ inputs.dry_run == false && inputs.action != 'delete' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep -v "$TAG" && \
+          git tag $TAG && git push origin $TAG
+    - name: check if tag has been pushed
+      if: ${{ inputs.dry_run == false && inputs.action == 'push' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep "$TAG"
+          git fetch -t
+          git tag -l "$TAG"
+    - name: delete if action==delete and tag exists in the crucible repo
+      if: ${{ inputs.dry_run == false && inputs.action == 'delete' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep "$TAG" && \
+          git push --delete origin $TAG
+    - name: check if tag has been deleted
+      if: ${{ inputs.dry_run == false && inputs.action == 'delete' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep -v "$TAG"
+
+  get-sub-projects:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      repos: ${{ steps.get-repos.outputs.repos }}
+    steps:
+    - name: Get the list of sub-projects
+      id: get-repos
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          # get first column (repo name) and add to a bash array ( a b c)
+          projects=( $(grep -v ^# config/default_subprojects | awk {'print $1'} ) )
+          # builtin implict join array a,b,c
+          printf -v list '%s,' "${projects[@]}"
+          # convert to a comma separated list [a,b,c]
+          echo "repos=[${list%,}]" >> $GITHUB_OUTPUT
+    - name: Get the sub-projects list
+      id: get-projects-list
+      env:
+          REPOS: ${{ needs.get-repos.outputs.repos }}
+      run: echo "$REPOS"
+
+  projects-tag:
+    needs:
+    - get-sub-projects
+    strategy:
+        matrix:
+            repository: ${{ needs.get-sub-projects.outputs.repos }}
+    steps:
+    - name: Display sub-project repository name
+      run: |
+          echo "repository=${{ matrix.repository }}"
+    - name: checkout sub-project repository
+      uses: actions/checkout@v4
+      with:
+        repository: perftool-incubator/${{ matrix.repository }}
+        ref: ''
+        path: ''
+    - name: get remote tags
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git ls-remote --tags origin
+    - name: create release tag
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git tag $TAG
+    - name: push release tag
+      if: ${{ inputs.dry_run == false && inputs.action == 'push' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git ls-remote --tags origin | grep -v "$TAG" && \
+          git tag $TAG && git push origin $TAG
+    - name: check if tag has been pushed
+      if: ${{ inputs.dry_run == false && inputs.action == 'push' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git ls-remote --tags origin | grep "$TAG"
+          git fetch -t
+          git tag -l "$TAG"
+    - name: delete if action==delete and tag exists in the project repo
+      if: ${{ inputs.dry_run == false && inputs.action == 'delete' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git ls-remote --tags origin | grep "$TAG" && \
+          git push --delete origin $TAG
+    - name: check if tag has been deleted
+      if: ${{ inputs.dry_run == false && inputs.action == 'delete' }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep -v "$TAG"
+
+  crucible-release-complete:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    needs:
+    - release-tag
+    - crucible-tag
+    - projects-tag
+    steps:
+    - name: Confirm Success
+      run: echo "crucible-release-complete"
--- a/spec/release-tag.md
+++ b/spec/release-tag.md
@@ -1,0 +1,69 @@
+# Release tag
+Implement a "Release" mechanism for installing "consistent" versions of
+Crucible.
+
+## Problem description
+We need to be able to install a "version" of Crucible that is "locked down"
+and won't change at all if it is installed over time.
+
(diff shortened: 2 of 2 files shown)


.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,125 @@
...
+  workflow_call:
+    inputs:
+      dry-run:
+        required: false
+        type: boolean
+        description: Do NOT push new generated tag to the upstream repositories
+        default: false
+
+env:
+  GITHUB_WORKSPACE: ${{ inputs.github_workspace }}
```
k-rister, line 13: `inputs.github_workspace` does not exist above

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,125 @@
...
+      run: |
+          year="$(date +%Y)"
+          month="$(date +%m)"
+          quarter="$(((month-1)/3+1))"
+          echo "tag=$year.$quarter" >> $GITHUB_OUTPUT
+    - name: Get release tag
+      id: get-release-tag
+      env:
+          TAG: ${{ needs.release-tag.outputs.tag }}
+      run: echo "$TAG"
```
k-rister, line 33: Unclear to me if this should be a separate job or not.

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,125 @@
...
+        repository: perftool-incubator/crucible
+        ref: master
+        path: crucible
+    - name: check if tag exists on the crucible repo
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep -v "$TAG"
```
k-rister, line 50: Any reason why you would not just run `git tag --list <tag>`?

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,125 @@
...
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          # get first column (repo name) and add to a bash array ( a b c)
+          projects=( $(grep -v ^# config/default_subprojects | awk {'print $1'} ) )
+          # builtin implict join array a,b,c
+          printf -v list '%s,' "${projects[@]}"
+          # convert to a comma separated list [a,b,c]
+          echo "repos=[${list%,}]" >> $GITHUB_OUTPUT
+
```
k-rister, line 68: I think you need a step/job here to display the output from the previous job (`sub-projects`) -- that could become very important when debugging.

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,125 @@
...
+        ref: ''
+        path: ''
+
+    - name: create release tag
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd ${{ matrix.repository }}
+          git tag $TAG
```
k-rister, line 95: Should there be a check to see if the tag already exists so that it is not overwritten?

.github/workflows/crucible-scheduled.yaml
```diff
@@ -16,3 +16,9 @@ jobs:
     secrets:
       production_registry_auth: ${{ secrets.CRUCIBLE_PRODUCTION_ENGINES_REGISTRY_AUTH }}
       quay_oauth_token: ${{ secrets.CRUCIBLE_QUAYIO_OAUTH_TOKEN }}
+
+  call-crucible-release:
+    uses: perftool-incubator/crucible/.github/workflows/crucible-release.yaml@main
+    needs: call-core-crucible-ci_scheduled
+    with:
+      github_workspace: "$GITHUB_WORKSPACE"
```
k-rister, line 24: You are adding this to a workflow that is echeduled to run monthly, but you have it designed to create quarterly versions.  So that would result in the version being set once (month 1 of the quarter) and then overwritten twice (months 2 & 3 of the quarter).  That seems wrong.  Shouldn't this be only run once per quarter?

k-rister:
I think we might want to do some path/path-ignore stuff to allow you to work on this without having to run the entire CI stack every time.

.github/workflows/crucible-release.yaml
rafaelfolco: good catch

.github/workflows/crucible-release.yaml
rafaelfolco: I'd keep it there until it is used somewhere else

.github/workflows/crucible-release.yaml
rafaelfolco: if you create a tag local, git tag is enough. We want to check the remote tags, after all, the push command may fail, or the tag might not be pulled from the repo depending on how you are pulling data.

.github/workflows/crucible-release.yaml
rafaelfolco: agreed, added.

.github/workflows/crucible-release.yaml
rafaelfolco: It depends on remote-tag job, and it remote-tag will fail if the tag already exists... this makes me re-think the way I am adding this dependency. I don't want the job to fail if the tag exists, so I am changing that logic.

.github/workflows/crucible-scheduled.yaml
rafaelfolco: Yeah, the original idea assumed the job would do nothing if the tag exists.... I am changing for an independent scheduled one.

rafaelfolco:
> I think we might want to do some path/path-ignore stuff to allow you to work on this without having to run the entire CI stack every time.

ack, will take a look

k-rister:
Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed?  I can see us need to backport a hotfix potentially and that would require a branch...

rafaelfolco:
> Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...



> Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...



> Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...

That's an interesting idea. Maintaining multiple branches may be problematic in many ways...
My suggestion would be only one single branch that points to the latest release, so we focus in one single backport. 
Say you have 2024.1, 2024.2, 2024.3, 2024.4, 2025.1, being 2025.1 the latest stable release. If you find a bug, so you'd only fix the latest stable branch, labeled as "stable".

k-rister:
> > Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...
> 
> That's an interesting idea. Maintaining multiple branches may be problematic in many ways... My suggestion would be only one single branch that points to the latest release, so we focus in one single backport. Say you have 2024.1, 2024.2, 2024.3, 2024.4, 2025.1, being 2025.1 the latest stable release. If you find a bug, so you'd only fix the latest stable branch, labeled as "stable".

I don't think we can really limit it to a single releast unfortunately.  If we have versions, and people are using those versions and they are experiencing a bug then we need to be able to fix it.  This is kinda the whole reason that maintaining versions is something we have avoided...but if we are going to have versions then I think this is a necessary evil.

If we have versions 2024.1, 2024.2, 2024.3 and 2024.4 and we find a bug in code that has been around awhile then I think it's very realistic to think that we are going to need hotfixes to each version and we will end up with 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 :(

I suppose we could also consider that we should create our versions as 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 so that if/when we make hot fixes that the versions are logically contiguous with the updates being 2024.1.2, 2024.2.2, 2024.3.2, and 2024.4.2.

@atheurer @jtaleric thoughts?

rafaelfolco:
> > > Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...
> > 
> > 
> > That's an interesting idea. Maintaining multiple branches may be problematic in many ways... My suggestion would be only one single branch that points to the latest release, so we focus in one single backport. Say you have 2024.1, 2024.2, 2024.3, 2024.4, 2025.1, being 2025.1 the latest stable release. If you find a bug, so you'd only fix the latest stable branch, labeled as "stable".
> 
> I don't think we can really limit it to a single releast unfortunately. If we have versions, and people are using those versions and they are experiencing a bug then we need to be able to fix it. This is kinda the whole reason that maintaining versions is something we have avoided...but if we are going to have versions then I think this is a necessary evil.
> 
> If we have versions 2024.1, 2024.2, 2024.3 and 2024.4 and we find a bug in code that has been around awhile then I think it's very realistic to think that we are going to need hotfixes to each version and we will end up with 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 :(
> 
> I suppose we could also consider that we should create our versions as 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 so that if/when we make hot fixes that the versions are logically contiguous with the updates being 2024.1.2, 2024.2.2, 2024.3.2, and 2024.4.2.
> 
> @atheurer @jtaleric thoughts?
IMHO we should simplify this process as much as we can. If we find a bug, we can fix in the main branch and in the stable branch, which points to the latest stable release. If users really need a specific bug fixed in an old release, update or do your own backport.

jtaleric:
> > > > Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...
> > > 
> > > 
> > > That's an interesting idea. Maintaining multiple branches may be problematic in many ways... My suggestion would be only one single branch that points to the latest release, so we focus in one single backport. Say you have 2024.1, 2024.2, 2024.3, 2024.4, 2025.1, being 2025.1 the latest stable release. If you find a bug, so you'd only fix the latest stable branch, labeled as "stable".
> > 
> > 
> > I don't think we can really limit it to a single releast unfortunately. If we have versions, and people are using those versions and they are experiencing a bug then we need to be able to fix it. This is kinda the whole reason that maintaining versions is something we have avoided...but if we are going to have versions then I think this is a necessary evil.
> > If we have versions 2024.1, 2024.2, 2024.3 and 2024.4 and we find a bug in code that has been around awhile then I think it's very realistic to think that we are going to need hotfixes to each version and we will end up with 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 :(
> > I suppose we could also consider that we should create our versions as 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 so that if/when we make hot fixes that the versions are logically contiguous with the updates being 2024.1.2, 2024.2.2, 2024.3.2, and 2024.4.2.
> > @atheurer @jtaleric thoughts?
> 
> IMHO we should simplify this process as much as we can. If we find a bug, we can fix in the main branch and in the stable branch, which points to the latest stable release. If users really need a specific bug fixed in an old release, update or do your own backport.

I have seen it both ways -- either branching or tags. I personally find tags easier to maintain, but that is due to my workflow. 

Where I think things with Crucible get complicated are tagging across the project.. For example, does `v0.0.1` (made up tag) of crucible work with `v0.0.1` of rickshaw?  Keeping the mapping clean might be problematic -- maybe I am overthinking it?

k-rister:
> > > > > Have you thought at all about using branches instead of tags, or at least being able to switch to a branch if needed? I can see us need to backport a hotfix potentially and that would require a branch...
> > > > 
> > > > 
> > > > That's an interesting idea. Maintaining multiple branches may be problematic in many ways... My suggestion would be only one single branch that points to the latest release, so we focus in one single backport. Say you have 2024.1, 2024.2, 2024.3, 2024.4, 2025.1, being 2025.1 the latest stable release. If you find a bug, so you'd only fix the latest stable branch, labeled as "stable".
> > > 
> > > 
> > > I don't think we can really limit it to a single releast unfortunately. If we have versions, and people are using those versions and they are experiencing a bug then we need to be able to fix it. This is kinda the whole reason that maintaining versions is something we have avoided...but if we are going to have versions then I think this is a necessary evil.
> > > If we have versions 2024.1, 2024.2, 2024.3 and 2024.4 and we find a bug in code that has been around awhile then I think it's very realistic to think that we are going to need hotfixes to each version and we will end up with 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 :(
> > > I suppose we could also consider that we should create our versions as 2024.1.1, 2024.2.1, 2024.3.1, and 2024.4.1 so that if/when we make hot fixes that the versions are logically contiguous with the updates being 2024.1.2, 2024.2.2, 2024.3.2, and 2024.4.2.
> > > @atheurer @jtaleric thoughts?
> > 
> > 
> > IMHO we should simplify this process as much as we can. If we find a bug, we can fix in the main branch and in the stable branch, which points to the latest stable release. If users really need a specific bug fixed in an old release, update or do your own backport.
> 
> I have seen it both ways -- either branching or tags. I personally find tags easier to maintain, but that is due to my workflow.
> 
> Where I think things with Crucible get complicated are tagging across the project.. For example, does `v0.0.1` (made up tag) of crucible work with `v0.0.1` of rickshaw? Keeping the mapping clean might be problematic -- maybe I am overthinking it?

@jtaleric *I think* we have the cross project part figured out.  The real question for you is what are your thoughts on backported hotfixes -- if in Novermber/December you are running Crucible version 2024.3 (ie. our 3rd quarter release) and we discover something is broken that you need a fix for, do you want to be forced to upgrade to the current release's hotfix (2024.4.x) or do you want that hotfix backported to the release you are on so that you can transition from 2024.3.x to 2024.3.x+1.  Based on the conversations we have had I would think you want 2024.3.x+1, but maybe I'm wrong.

My current line of thinking is that we would "support" the last year's worth of releases -- so basically there would always be 4 supported versions available since we are currently planning a quarterly release schedule.  The only backports would be for critical hotfixes -- if a user wants new/added functionality they either need to upgrade to one of the newer releases or move to the bleeding edge which is HEAD of master/main (ie. what everyone is running right now -- assuming they update in a timely fashion).  The question is would the hotfixes be applied to all 4 releases or just the most recent one.  I'm of the opinion that it doesn't make sense to "support" releases if we aren't going to fully support them with backported hotfixes.

FWIW, I sorta came up with the year of "support" when I put together the prebuilt image stuff.  As currently configured all the prebuilt images are valid for at least a year, if not longer (they will be valid as long as they are the most recent version of the image and/or they are actively being used with the expiration timestamp refresh stuff enabled).

k-rister (changes requested):
I think we should probably also create a complimentary workflow that would allow us to delete a release.

With the changes I proposed here and the additional workflow we would be able to very easily create and delete releases for testing purposes.

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,145 @@
...
+    outputs:
+      tag: ${{ steps.gen-release-tag.outputs.tag }}
+    steps:
+    - name: Generate release tag
+      id: gen-release-tag
+      run: |
+          year="$(date +%Y)"
+          month="$(date +%m)"
+          quarter="$(((month-1)/3+1))"
+          echo "tag=$year.$quarter" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
rafaelfolco: ack

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,145 @@
...
+      if: ${{ inputs.dry_run == false }}
+      env:
+        TAG: ${{ needs.release-tag.outputs.tag }}
+      run: |
+          cd $GITHUB_WORKSPACE
+          cd crucible
+          git ls-remote --tags origin | grep -v "$TAG" && \
+          git tag $TAG && git push origin $TAG
+
+  sub-projects:
```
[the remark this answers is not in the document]
rafaelfolco: +1

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,145 @@
...
+on:
+  schedule:
+    - cron: "0 0 1 */3 *"
+    paths-ignore:
+    - LICENSE
+    - '**.md'
+    - .github/workflows/crucible-scheduled.yaml
+    - .github/workflows/faux-crucible-ci.yaml
+    - .github/workflows/faux-ci.yaml
+    - 'docs/**'
```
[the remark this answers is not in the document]
rafaelfolco: Right

.github/workflows/crucible-release.yaml
```diff
@@ -0,0 +1,145 @@
...
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        required: false
+        type: boolean
+        description: Do NOT push new generated tag to the upstream repositories
+        default: false
```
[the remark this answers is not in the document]
k-rister: I would just say lets add custom versions as the next step because I think that will facilitate better testing as you do the other stuff.

k-rister (approved):

PR #358, opened 2024-07-23
Fix crucible-release workflow

rafaelfolco:
Required property is missing: runs-on

Adjust timeout values

### DIFF
--- a/.github/workflows/crucible-release.yaml
+++ b/.github/workflows/crucible-release.yaml
@@ -26,7 +26,7 @@ on:
 jobs:
   release-tag:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


rafaelfolco:
https://github.com/perftool-incubator/crucible/actions/runs/10060253294
`
[Invalid workflow file: .github/workflows/crucible-release.yaml#L130](https://github.com/perftool-incubator/crucible/actions/runs/10060253294/workflow)
The workflow is not valid. .github/workflows/crucible-release.yaml (Line: 130, Col: 5): Required property is missing: runs-on
`

k-rister (approved):
Is it really that slow?

PR #384, opened 2024-08-15
Locked-down installs of crucible

rafaelfolco:
Add option for a locked-down version of Crucible.
Includes a --list-releases option to show tags.
Crucible installer --release <TAG> installs from
a <TAG>, or selects from a list if <TAG> is omitted (interactive mode).

Only default git repo is supported, and git branch cannot be specified if --release <TAG> option is used.


rafaelfolco:
working on tests and doc

k-rister (approved):

PR #388, opened 2024-08-20
add spec file detailing plans for a registries.json config file

k-rister:


k-rister:
Let's merge a different PR first: https://github.com/perftool-incubator/crucible/pull/384

In order to avoid useless CI re-runs...this PR just runs the faux jobs so it can be rebased and pass CI again really easily.

atheurer (commented):
Does /etc/sysconfig/crucible need to have a line to point to the registries.json?

k-rister:
> Does /etc/sysconfig/crucible need to have a line to point to the registries.json?
> 

I think something that we need to decide is where this file should be located and that would determine if it needs a pointer in /etc/sysconfig/crucible.

We could place it in ~/.crucible and then we would be able find it automatically.

We could place it "out of tree" somewhere else and then probably we would have to have a pointer in /etc/sysconfig/crucible.

We could also place it "in tree" and be able to find it very easily, but that might come with some fast forward/merge problems if/when we need to make changes/additions to the file down the road.  Since this file will be full of "local" data and only the basic structure would be known/controlled by upstream I'm not sure that "in tree" makes sense -- it might end up being more trouble than it's worth.

atheurer (approved):

PR #394, opened 2024-08-30
Fix delete branch/tag when does not exist

rafaelfolco:
Ignore error if attempt to delete heads or tags that don't exist. Fix minor issue with the tag listing.


rafaelfolco:
won't merge this on a Friday

k-rister (approved):

PR #395, opened 2024-09-02
Fix force option for delete tag

rafaelfolco:
error: src refspec || does not match any


rafaelfolco:
https://github.com/perftool-incubator/crucible/actions/runs/10668543448/job/29568517058

k-rister (approved):

PR #397, opened 2024-09-03
activate repos.json

k-rister:


k-rister:
going to experiment with using the CI environment to flush out all the bugs in this PR...we shall see how this goes

PR #398, opened 2024-09-03
activate repos.json

k-rister:


crucible-install.sh
```diff
@@ -291,6 +293,71 @@ function select_release {
...
+                update_target=1
+
+                # what do we do about the mode?
+
+                if [ "${BRANCH}" == "HEAD" ]; then
+                    echo "Detected CI environment, setting crucible primary-branch to '${BRANCH}' in '${REPO_FILE}'" >> ${GIT_INSTALL_LOG}
+                    jq_update ${REPO_FILE} primary-branch --arg primary_branch ${BRANCH} '(.official[] | select(.name == "crucible") | ."primary-branch") |= $primary_branch'
+                fi
+            fi
+        fi
```
k-rister, line 346: Still need to sort out a couple of questions here.

k-rister:
This is almost complete, just need to sort out a few details with input from others (@rafaelfolco , @atheurer) before taking this out of draft mode and letting the full CI run.

k-rister:
This PR is ready for (final) review.

tests/test-installer
```diff
@@ -62,68 +111,98 @@ for arg_mode in client-server engine; do
...
     https://github.com/perftool-incubator/crucible.git \
     | awk -F/ '{print$NF}')
+stop_test
 
+start_test
 ## show releases, test rc and stdout
 stdout=$(sudo ./crucible-install.sh \
+    --git-repo ${CRUCIBLE_DIR} \
     --list-releases)
 test "$?" = "0" || exit 1
```
rafaelfolco, line 173: I like the start/stop idea, wondering if we can improve how the expected output is checked at the end of each test...

bin/_update-git
```diff
@@ -12,55 +12,106 @@ exec 2>&1
...
 
-	if ! echo "${stash_output}" | grep -q "No local changes to save"; then
-            # reapply local changes
-            if ! git stash pop > /dev/null; then
-		echo -e "${stash_output}"
-		echo "ERROR: could not git-stash pop"
+    if [ "${PROJECT_CHECKOUT_MODE}" == "locked" ]; then
+	# return to the previous ref if is supposed to be locked there
+	echo "Returning to the previously checked out ref since I am in locked mode"
+	git checkout ${current_ref}
```
rafaelfolco, line 95: OK, now it makes more sense to me. The action depends on the mode, and this is not about what we do with the mode.

bin/repo
```diff
@@ -199,21 +202,75 @@ function handle_scope() {
...
+            show)
+                config_show
+                ;;
+            help)
+                config_help
+                ;;
+            *)
+                config_help
+                exit 1
+                ;;
```
rafaelfolco, line 273: we can only show config now... I am just wondering if allowing the user to change the config in the future is a good idea. Only advanced users need to change the config, so they can change the json directly?

rafaelfolco (approved):
lgtm

tests/test-installer
k-rister: Yeah, it's an improvement but I feel like it could still be better.  For someone like me who isn't as familiar as looking at this job as you probably are it was quite painful at first just to figure out which test case was actually failing so this was an attempt to fix that.

bin/repo
k-rister: Well, I think it is complicated.  I do think we will probably want the ability to switch from one version to the next at a minimum.  So say for example you install a controller on version 2024.3 and that is in your repos.json but then eventually you want to move to 2024.4 when it is available.  I think we need a command for that...something like:

```
crucible repo config set-version 2024.4
```

and it would ripple through and change all of the repos for you individually.  Similarly, once you do that you might find something that breaks your workflow and so you might want to go backwards:

```
crucible repo config set-version 2024.3
```

That is very easy and I don't think a "power user" action.

Beyond that, it very well be that other actions I had in mind are optional since they are possible to do by editing the config file directly.  However, it should be noted that scripting them with a nice interface does afford us the ability to do things like validate the inputs, check the return codes for the changes of individual settings, etc. that a manual edit wouldn't do -- basically it would give us greater ability to maintain the integrity of the config file.

Additionally, while I wrote a schema for this JSON file at the moment I'm not actually doing anything with it.  I figured I would add that later.  Validation of the file and the ability to revert (or something else) is the kind of thing we can do when the changes are scripted as opposed to when manual.  Another reason why I didn't implement validation yet is I need to figure out when / how often to validate -- I don't want to do it over and over again needlessly so I need to carefully determine where to add calls to that.

k-rister:
Waiting to merge this until Monday, Sept. 23, 2024 in case it causes any troubles that need to be dealt with.

PR #400, opened 2024-09-23
make the git repo used by the installer consistent with other aspects…

k-rister:
… of crucible by dropping '.git'


rafaelfolco (approved):
lgtm

PR #415, opened 2024-10-21
Dev kmr

k-rister:


bin/jqlib
```diff
@@ -16,6 +16,14 @@ function jq_update() {
...
     fi
 
+    diff ${FILE} ${TMP_FILE} > /dev/null
+    DIFF_RC=$?
+    if [ ${DIFF_RC} -eq 0 ]; then
+        exit_error "Failed to update ${REASON} in ${FILE} -- no changes detected"
+    elif [ ${DIFF_RC} -gt 1 ]; then
+        exit_error "Failed to update ${REASON} in ${FILE} -- an error occurred"
+    fi
+
```
rafaelfolco, line 26: in what situation do we use this update function ?

k-rister: Right now it's only when we are manipulating the `repos.json` file, but going forward it will also be used to manipulate the `registries.json` file that I'm starting to work on.

k-rister: I should note that I added this additional detection logic when I noticed that one of the queries I had written to update `repos.json` was failing silently -- I had written a syntactically correct query so jq was returning zero but the query itself was failing to match a specific record so nothing was getting updated.

PR #420, opened 2024-10-31
Adding services.json.md

jtaleric:
Adding proposal for OpenSearch in services.json


k-rister (approved):
Seems like a good starting point to me.

jtaleric:
ack - any other feedback or can we merge?

k-rister:
Let's merge and then hopefully we can solicit some feedback.

PR #424, opened 2024-11-14
check cdmver before getting period

atheurer:


k-rister (approved):

k-rister:
@atheurer To be clear, this PR should allow your existing `v8dev` PR to pass CI, but I'm wondering if we should also include embedding of the CDM version in the documents as part of the new `v8dev` changes.

atheurer:
> @atheurer To be clear, this PR should allow your existing `v8dev` PR to pass CI, but I'm wondering if we should also include embedding of the CDM version in the documents as part of the new `v8dev` changes.

Yes, I was thinking about this, too.  I do think we need to embed the version in the docs, but I wanted to make that a different effort if possible.  We probably need to pull the logic from rickshaw-index that finds the actual cdm version present in the indices and put that in toolbox or maybe just a script in CDM.

k-rister:
> > @atheurer To be clear, this PR should allow your existing `v8dev` PR to pass CI, but I'm wondering if we should also include embedding of the CDM version in the documents as part of the new `v8dev` changes.
> 
> Yes, I was thinking about this, too. I do think we need to embed the version in the docs, but I wanted to make that a different effort if possible. We probably need to pull the logic from rickshaw-index that finds the actual cdm version present in the indices and put that in toolbox or maybe just a script in CDM.

I guess that poses an interesting question, does making it part of a different effort mean that it is part of a newer version (`v9dev`)?  It seems like unless we make the changes in an atomic fashion that each change means a new version, and switching versions seems to be annoying at best.  But if you are comfortable with that then I'm fine with it.

PR #435, opened 2025-01-10
Add caller to crucible release workflow

rafaelfolco:
Crucible release workflow on Schedule needs a stub caller to set input variables as they only work with workflow_dispatch, which is manual trigger.

### DIFF
--- a/.github/workflows/crucible-release-schedule.yaml
+++ b/.github/workflows/crucible-release-schedule.yaml
@@ -1,0 +1,24 @@
+---
+name: crucible-release-schedule
+on:    # yamllint disable-line rule:truthy
+  schedule:
+    # Runs every 3 months (1st day of the quarter)
+    - cron: "0 0 1 */3 *"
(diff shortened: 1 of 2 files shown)


.github/workflows/crucible-release-schedule.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  call-crucible-release-schedule:
+    uses: ./.github/workflows/crucible-release.yaml
+    with:
+      action: "push"
+    secrets:
+      PRIVATE_KEY__TAG_CRUCIBLE_RELEASE: ${{ secrets.PRIVATE_KEY__TAG_CRUCIBLE_RELEASE }}
+
+  # TO-DO: add a job to remove old tags (2years?)
```
k-rister, line 16: I think the plan has always been 1 year of support for each release.

k-rister (approved):

PR #440, opened 2025-02-10
switch from tags to branches for release listing logic

k-rister:


tests/test-installer
```diff
@@ -212,8 +212,8 @@ done
 
 start_test
 # RELEASE: --release and --list-releases
-## prep step: check tags in remote repo
-tags=$(git ls-remote --tags \
+## prep step: check branches in remote repo
+tags=$(git ls-remote --heads \
```
rafaelfolco, line 216: rename tags var?

rafaelfolco (changes requested):

tests/test-installer
k-rister: Good call.

rafaelfolco (approved):

PR #452, opened 2025-03-18
Allow Configuring Random Valkey Password

drewrip:
Creates functionality to generate a random password for valkey. This password is then passed to rickshaw via a new cli argument that is included in the following rickshaw PR: https://github.com/perftool-incubator/rickshaw/pull/607.

### DIFF
--- a/bin/base
+++ b/bin/base
@@ -165,6 +165,12 @@ function close_firewall_port() {
     esac
 }
 
+function generate_rand_pass() {
+    local output_file
+    output_file="$1"
+    head -c 4096 /dev/urandom | sha256sum | awk '{print $1}' > $output_file
+}
+
 function start_httpd() {
     local RC httpd_cmd
     httpd_cmd="/usr/sbin/httpd -DFOREGROUND"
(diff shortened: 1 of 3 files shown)


bin/base
```diff
@@ -200,7 +206,8 @@ function stop_httpd() {
 
 function start_valkey() {
     local RC valkey_cmd
-    valkey_cmd="valkey-server /etc/valkey/valkey.conf"
+    generate_rand_pass /tmp/valkey_pass
+    valkey_cmd="valkey-server --requirepass=$(cat /tmp/valkey_pass) /etc/valkey/valkey.conf"
```
k-rister, line 210: I think you are going to want to move these two lines inside the if statement otherwise every time it checks for valkey (ie. every run) it will be resetting the persistently stored password without restarting the service.

bin/base
```diff
@@ -218,6 +225,7 @@ function start_valkey() {
 function stop_valkey() {
     local RC=0
     echo -n "Checking for valkey"
+    rm -f /tmp/valkey_pass
```
k-rister, line 228: This one is less important, but it seems like it would be "cleaner" to put this inside the if statement also.

k-rister (changes requested):
A few comments and I went ahead and canceled this PR's CI run since it needs to wait to run until after the other PR is merged.

bin/base
drewrip: Agree. Fixed!

k-rister:
@drewrip I did some debugging and found a couple minor issues that I pushed a fix for.  It looks like it's making progress through CI now.

k-rister (approved):

PR #465, opened 2025-06-04
change the controller image to be based on Fedora 42 instead of Fedora 41

k-rister:


k-rister:
Something appears to be wrong with OpenSearch.  Debugging...

PR #468, opened 2025-06-27
limit the supported releases to the 4 most recent quarterly generated…

k-rister:
… versions


k-rister:
This PR is important because it will limit the amount of testing is done when we have more than 4 releases generated which start happening July 1st (next week).

PR #470, opened 2025-07-11
Manage opensearch instances for crucible

atheurer:
- Soon crucible wil be able to use multiple opensearch instances.  This code manages a json file to track those instances, much like 'crucible repo' and 'crucible registry' do.
- This commit does not yet enable multiple instance usage.  A different commit will do that, where before calling get-result-summary.js and get-metric-data.js a call to manage_instances will be made to find instances to query-from and get the correct query-option that is passed to get-metric-data/get-result-summary

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -40,6 +40,15 @@ elif [ "${1}" == "ls" -o "${1}" == "tags" -o "${1}" == "result-completion" ]; th
             EXIT_VAL=$?
             ;;
     esac
+elif [ "${1}" == "instances" ]; then
+    shift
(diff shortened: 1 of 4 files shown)


atheurer:
> Should something be invoking `test_instances.py`?

At some point, yes, but was not sure if you wanted it to be included as part of a CI test

k-rister:
@rafaelfolco Any guidance?  `ci.yaml` is your creation and Python unit testing is also your area of expertise (relative to me anyway).

k-rister (approved):

PR #479, opened 2025-08-08
Multi opensearch 1

atheurer:


atheurer:
Hoping the CI will run with the recent change to CDM

k-rister (changes requested):
appears to still be broken

k-rister (approved):

atheurer:
This is going to require a new way to get the primary period IDs.  That work will be done with a new script in CDM instead of trying to extract from local files

PR #486, opened 2025-08-26
add special case handling to subprojects-install for crucible-ci targ…

k-rister:
…et repositories

- if the repository value is 'CRUCIBLE-CI' then subprojects-install will not alter the subproject's configuration in order to avoid breaking the special CI configuration


k-rister:
This PR is part of the solution to https://github.com/perftool-incubator/crucible-ci/issues/176

PR #578, opened 2026-05-10
feat: add --set-release to installer for switching between releases

k-rister:
## Summary
- Add `--set-release` option to `crucible-install.sh` for switching an existing installation between quarterly releases or back to upstream
- Preserves unofficial repos and custom repository URLs across release switches
- Stops crucible services before switching, rolls back on failure
- Includes 14 test cases in `tests/test-installer`

## Details
- `--set-release <release>` — switch to a quarterly release (locked mode)
- `--set-release upstream` — switch back to following primary branches
- `--set-release select` — interactive release selection
- `--set-release-force` — allow unsupported (older) releases

### Why the installer?
Switching releases checks out a different branch of the crucible repo, which
overwrites all repo-managed scripts. If the target release predates the
set-release feature, the command would not exist after the switch, leaving the
user unable to switch again. While the installer does live in the
release-managed code, it is a well established practice to manually download it
from the upstream repo (e.g., via curl/wget) without using a git clone
operation. This means a user can always acquire a current version of the
installer with the set-release feature regardless of which release is currently
checked out.

## Test plan
- [ ] CI passes test-installer (install job)
- [ ] Manual round-trip: upstream → release → different release → upstream

### DIFF
--- a/crucible-install.sh
+++ b/crucible-install.sh
@@ -16,6 +16,8 @@ DEFAULT_QUAY_EXPIRATION_LENGTH="13w"
 GIT_REPO=""
 GIT_BRANCH=""
 GIT_TAG=""
+SET_RELEASE=""
+SET_RELEASE_FORCE=0
 VERBOSE=0
 
 # User Exit Codes
@@ -38,6 +40,9 @@ EC_RELEASE_DEFAULT_REPO_ONLY=18
 EC_RELEASE_CONFLICTS_WITH_BRANCH=19
 EC_INVALID_QUAY_EXPIRATION_LENGTH=20
 EC_OAUTH_FILE_NOT_FOUND=21
+EC_SET_RELEASE_NO_INSTALL=22
+EC_SET_RELEASE_NOT_FOUND=23
+EC_SET_RELEASE_CHECKOUT_FAIL=24
 
 # remove a previous installation log
 if [ -e ${GIT_INSTALL_LOG} ]; then
@@ -150,11 +155,12 @@ function usage {
     Crucible installer script.
 
     Usage: $0 --engine-registry <value> [ opt ]
+           $0 --set-release <release|upstream|select> [--set-release-force]
+
+    Install mode (new installation):
 
         --engine-registry <full registry url>
-        Engine registry.
-
-    optional:
+        Engine registry (required for install).
 
         --quay-engine-expiration-refresh-token <authentication file>>
         Quay OAuth authentication token file for refreshing engine image expiration timestamps.
@@ -167,7 +173,7 @@ function usage {
 
         --engine-auth-file <authentication file>
         Authentication file for pushing images to the remote registry.
-  
+
         --engine-tls-verify true|false
         Use TLS verification.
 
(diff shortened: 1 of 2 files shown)


crucible-install.sh
```diff
@@ -450,14 +786,23 @@ while true; do
...
 	    shift;
-	    list_releases
+	    list_supported_releases
 	    exit
 	    ;;
 	--release)
 	    shift;
 	    select_release "$1"
 	    shift;
 	    ;;
```
rafaelfolco, line 796: --release is still relevant ?

rafaelfolco (approved):
I'm just confused about the --release arg.

crucible-install.sh
k-rister: Yes, `--release` is the install time option whereas `--set-release` is used to configure the release for an existing install.  Nothing is changing from an install time perspective.  This is entirely about adding new functionality for post-install configuration.

This has long been on my todo list but I finally needed it late last week and here is the scenario.  I was working on an update to the controller image and when that was being tested in CI it kept failing on the 2025.3 release and Claude and I were unable to debug it looking at CI runs retroactively.

Thus this feature is born.  This allows you to take a working installation and convert it from release to another -- in my case `upstream` to `2025.3` in order to debug the problem locally in that environment.  However, this is mainly a development workflow.

The real user workflow usage for this is the following:  A user installs on a release, say 2026.1 and uses that for many months.  Eventually they need a feature that is available in a newer release, say 2026.2, but without this feature their only options are 1) reinstall on the new release or 2) manually change every single repo's release (note that reinstalling could lose local customizations and/or configuration changes).  With this capability you just run: `./crucible-install.sh --set-release 2026.2` and away you go.

There is some added complexity in the actual code flow to handle things like modified repo URLs (a development environment need) and external repos (since they aren't tracked as part of the releases).

Hopefully the `--help` output makes segmentation obvious:

```

    Crucible installer script.

    Usage: ./crucible-install.sh --engine-registry <value> [ opt ]
           ./crucible-install.sh --set-release <release|upstream|select> [--set-release-force]

    Install mode (new installation):

        --engine-registry <full registry url>
        Engine registry (required for install).

        --quay-engine-expiration-refresh-token <authentication file>>
        Quay OAuth authentication token file for refreshing engine image expiration timestamps.

        --quay-engine-expiration-refresh-api-url <api url>
        Quay API URL that is used to operate on the engine registry.

        --quay-engine-expiration-length <length>
        How long should a Quay repo allow the engine images to live before they expire.

        --engine-auth-file <authentication file>
        Authentication file for pushing images to the remote registry.

        --engine-tls-verify true|false
        Use TLS verification.

        --controller-registry <full registry url>
        Controller registry (default is quay.io/crucible/controller:latest).

        --name <your full name>
        --email <your email address>
        Name and email of Crucible user.

        --git-repo <repo path/url>
        --git-branch <branch name>
        Repository and branch to install Crucible from.

        --release [<release>|select]
        Release tag <release> to install a locked-down version of Crucible.
        The option '--release select' lists all available releases on
        interactive prompt.

        --verbose
        Verbose mode that provides more debugging info.

    Release management (existing installation):

        --set-release <release|upstream|select>
        Switch an existing installation to a specific quarterly release or
        back to upstream. Use 'select' for interactive release selection.

        --set-release-force
        Allow switching to unsupported (older) releases.
        Use with --set-release.

    General:

        --list-releases
        Show available releases from the remote repository.

    --help
    Displays this usage output.
```

PR #584, opened 2026-05-19
feat: add per-architecture image sourcing service support

k-rister:
## Summary

- Migrate `image-sourcing` in `services.json` from a single service to a per-architecture services map, enabling multi-arch image builds via remote native builders
- Each architecture gets its own service entry with independent `start`/`location` configuration
- `bin/_main` writes a per-arch URL mapping file (`image-sourcing-urls.json`) to the run config directory for rickshaw consumption
- Non-native `start: true` entries are fatal — local cross-arch builds are not supported due to buildah/QEMU incompatibilities; non-native architectures must use a remote service
- New `bin/_migrate-services-config` script handles migration from the old flat format and resolves the `default` placeholder key to the host's native architecture; runs on install, update, and every command invocation
- Legacy `--source-images-service-url` CLI arg is still passed to rickshaw for backward compatibility during the transition period

## Key files

| File | Change |
|---|---|
| `config/services.json` | Per-arch services map with `default` placeholder key |
| `schema/services.json` | Updated schema with `patternProperties` for arch keys |
| `bin/base` | Multi-arch `start/stop_image_sourcing()`, fatal error for non-native start |
| `bin/_main` | Build per-arch URL mapping file, check `start_image_sourcing` return code |
| `bin/_migrate-services-config` | New — idempotent config migration script |
| `bin/update` | Call migration during `crucible update` |
| `crucible-install.sh` | Call migration after install |

## Multi-arch services.json example

```json
{
    "image-sourcing": {
        "use": true,
        "services": {
            "x86_64": {
                "start": true,
                "location": {
                    "address": "localhost",
                    "port": 8888,
                    "protocol": "http"
                }
            },
            "aarch64": {
                "start": false,
                "location": {
                    "address": "arm-builder.example.com",
                    "port": 8888,
                    "protocol": "http"
                }
            }
        }
    }
}
```

## Test plan

- [ ] Fresh install: verify `default` key is resolved to native arch by migration
- [ ] `crucible update`: verify migration runs and resolves `default` key
- [ ] Single-arch run: verify backward-compatible behavior (single service, `image-sourcing-urls.json` written)
- [ ] Multi-arch config with remote builder: verify both URLs appear in `image-sourcing-urls.json`
- [ ] Non-native `start: true`: verify fatal error with clear message
- [ ] Old flat-format `services.json`: verify migration converts correctly

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,6 +112,26 @@ Follow existing code style conventions exactly — check surrounding code for pa
 - `~/.crucible/identity`: User identity (CRUCIBLE_NAME, CRUCIBLE_EMAIL)
 - Run data stored in `/var/lib/crucible/run/`
 
+## Multi-Architecture Image Sourcing
+
+Engine container images are built by the image sourcing service (SIS). `config/services.json` maps each CPU architecture to its own SIS instance under `image-sourcing.services`:
+
+```json
+{
+    "image-sourcing": {
+        "use": true,
+        "services": {
+            "x86_64": { "start": true, "location": { "address": "localhost", "port": 8888, "protocol": "http" } },
+            "aarch64": { "start": false, "location": { "address": "arm-builder.example.com", "port": 8888, "protocol": "http" } }
+        }
+    }
+}
+```
+
+- The repo ships with a `"default"` placeholder key; `bin/_migrate-services-config` resolves it to the host's native architecture (via `uname -m`) on install, update, and every command invocation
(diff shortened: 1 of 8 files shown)


atheurer (approved):
I wonder how we might handle something like a specific family of x86_64, like the v3 variant.  I think the build service stays the same, as long as they can compile to a different micro-arch than what they have.  I suppose this type of enhancement is more in the workshop domain.

k-rister:
> I wonder how we might handle something like a specific family of x86_64, like the v3 variant.  I think the build service stays the same, as long as they can compile to a different micro-arch than what they have.  I suppose this type of enhancement is more in the workshop domain.

Yeah, based off my experience with 'podman run --platform' I would say that would definitely have to be addressed in workshop.

See https://github.com/containers/buildah/discussions/5322

PR #589, opened 2026-05-21
fix: handle primary repo type in repo config update checkout

k-rister:
## Summary
- `crucible repo config update name=crucible checkout-target=<branch>` failed because the checkout path was constructed as `subprojects/primary/crucible`, which doesn't exist — the crucible repo lives at `CRUCIBLE_HOME` directly
- Special-case repos with type "primary" to use `CRUCIBLE_HOME` as the checkout directory

## Test plan
- [x] `crucible repo config update name=crucible checkout-target=<branch>` successfully checks out the branch
- [x] `crucible repo config update name=crucible checkout-target=master` switches back
- [x] Subproject checkout-target updates still work (e.g., `name=rickshaw`)


rafaelfolco (approved):
lgtm

PR #590, opened 2026-05-21
feat: add automatic port searching for service startup

k-rister:
## Summary
- Removes an invalid `local` declaration outside a function in `bin/_main`
- Adds `find_available_port()` helper to `bin/base` that uses `ss -tln` to detect port conflicts and automatically searches for the next available port (up to 100 attempts)
- Integrates port searching into `start_httpd()`, `start_cdm_server()`, and `start_image_sourcing()` — when a configured port is in use, the service starts on the next available port and `services.json` is updated so all components use the correct value
- Valkey and OpenSearch are out of scope — their ports are hardcoded across multiple repos and require a separate cross-repo effort

## Test plan
- [x] Block port 8080 with a socket listener, run `crucible start httpd` — finds 8081, logs WARNING, updates services.json
- [x] Verify `find_available_port` returns immediately on a free port with no warning
- [ ] Block port 3000, test `crucible start cdm-server`
- [x] Block port 8888, test `crucible start image-sourcing`
- [ ] Full `crucible run` with a port conflict on httpd

### DIFF
--- a/bin/_main
+++ b/bin/_main
@@ -372,7 +372,6 @@ elif [ "${1}" == "run" ]; then
         fi
 
         # Build per-architecture URL mapping and write to run config
-        local urls_json arch arch_address arch_port arch_protocol arch_url
         urls_json="{"
 
         for arch in $(jq_query ${SERVICES_CFG} '."image-sourcing".services | keys[]'); do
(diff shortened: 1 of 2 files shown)


HughNhan (approved):
A note to mysel. It searches the next available higher port number. 
LGTM

PR #629, opened 2026-07-01
feat: add remote archive storage via rclone

k-rister:
## Summary

Add the ability to upload and download result archives to/from remote S3-compatible storage via rclone. This extends the existing `crucible archive` and `crucible unarchive` commands with `--local`/`--remote` flags for flexible data replication, and adds `crucible archive config` for managing remote storage backends.

**Feature commits:**
- Add `remote-archive` configuration section to `services.json` with schema, migration, and documentation
- Add `crucible archive config` management command (add/remove/info/default)
- Add rclone backend translation layer (`bin/_remote_archive`) for S3 with tls-verify support
- Integrate `--local`/`--remote`/`--delete-local`/`--delete-remote` flags into archive and unarchive
- Extend `crucible ls --type archive --remote <name>` for remote listing

**Bugfix commits (pre-existing issues found during testing):**
- Filter non-directory entries from run result listing
- Use bare names instead of full paths in tab completion
- Strip carriage returns from podman container output in tab completion

**Key design decisions:**
- No flags = original move semantics (backward compatible); any flag = copy semantics (source preserved)
- `--delete-local` / `--delete-remote` for explicit opt-in move behavior after successful transfer
- Skip upload when archive already exists on remote
- Credentials stored in external files, referenced by path (same pattern as registries.json)
- rclone executes inside the controller container (installed via crucible#628)
- Remote management lives under `crucible archive config` (parallel to `crucible repo config`)

**Tested against:** rustfs S3-compatible server with archive, unarchive, ls, wildcard operations, skip-existing, copy/move semantics, and error handling.

## Test plan

- [ ] CI passes
- [ ] Manual: `crucible archive config add/remove/info/default` operations
- [ ] Manual: `crucible archive <dir> --remote <name>` (upload, source preserved)
- [ ] Manual: `crucible archive <dir> --local --remote <name>` (both, source preserved)
- [ ] Manual: `crucible archive <dir> --remote <name> --delete-local` (upload, source deleted)
- [ ] Manual: `crucible unarchive <name> --remote <name>` (download + extract)
- [ ] Manual: `crucible unarchive <name> --remote <name> --delete-remote` (download, remote deleted)
- [ ] Manual: `crucible ls --type archive --remote <name>` (remote listing)
- [ ] Manual: `crucible archive --help` / `crucible help archive`
- [ ] Manual: Tab completion for archive/unarchive commands
- [ ] Manual: No-flags archive/unarchive backward compatibility


config/services.json
```diff
@@ -37,6 +37,21 @@
...
+                "bucket": "crucible-archives",
+                "tls-verify": false,
+                "settings": {
+                    "endpoint": "http://192.168.12.100:9000",
+                    "region": "us-east-1"
+                }
+            }
+        },
+        "default": "test-s3"
+    },
```
k-rister: This is supposed to be empty.  It was accidentally committed with testing credentials in place.

k-rister: Will fix.

k-rister: Fixed.

atheurer:
some feedback from claude:

`● Review: PR #629 — feat: add remote archive storage via rclone

  This PR adds rclone-based S3 remote archive storage to crucible's archive/unarchive commands, with a new crucible archive config management interface, tab completion updates, and several bugfixes (bare names in completion, \r
  stripping, non-directory filtering). It's well-structured overall — config schema, migration, docs, and help text are thorough. Here are the findings that survived verification:

  ---
  Correctness

  1. bin/base:~921 — unarchive_run() --remote mode never removes the downloaded archive after extraction

  The comment says "download, extract to run dir, remove local copy" but no rm of ${ARCHIVE_FILE} appears after successful extraction. The local-only path (further down in the same function) does remove the archive after extraction
  (lines ~989-994). Each crucible unarchive <name> --remote <name> leaves an orphaned .tar.xz in the archive directory.

  2. bin/base:~1148 — Temp archive silently deleted on upload failure, losing minutes of compression work

  When --remote is used without --local, the archive is created in a mktemp -d directory. If remote_archive_upload fails (network timeout, expired credentials, S3 throttling), the temp directory is immediately rm -rf'd. For
  multi-gigabyte runs, xz --threads=0 compression can take minutes — that work is lost with no recovery path. The user must re-compress from scratch to retry. Consider preserving the staged archive on failure and printing its path.

  3. bin/_remote_archive:~131 — Skip-existing check silently swallows rclone errors

  remote_archive_upload checks for an existing archive with:
  existing=$(_run_rclone lsf ... 2>/dev/null | grep -Fx "${archive_basename}")
  If rclone lsf fails (auth error, unreachable endpoint, wrong bucket), stderr is suppressed, existing is empty, and the code falls through to attempt the upload. The user gets a confusing "Failed to upload" error with no indication that
  the existence check itself failed. Remove 2>/dev/null or check the _run_rclone exit code separately.

  Architecture

  4. bin/result-processor.py:~505 — Python ls --remote runs rclone on the host; all other remote operations run rclone in the container

  remote_archives_ls() calls subprocess.run(["rclone", "lsf", ...]) directly on the host, while _run_rclone() in bin/_remote_archive runs rclone inside the controller container. This means crucible ls --type archive --remote X requires
  rclone installed on the host, while crucible archive --remote X only needs rclone in the container image (PR #628). A user who sets up remote archives will find that archive/unarchive work but ls --remote fails with "rclone is not
  installed." The host-side invocation also exposes S3 credentials in ps aux to all host users.

  5. bin/result-processor.py:~519 — Duplicate rclone arg-building logic in Python and bash

  _build_rclone_args_py() reimplements the same config-to-CLI-flags transformation as _build_rclone_args() in bash. Adding a new backend type or S3 setting requires updating both, and divergence will cause ls --remote to behave
  differently from archive --remote. This is closely related to finding #4 — unifying the execution model would eliminate the duplication.

  Style

  6. bin/_main:~636 — Unarchive else-block content not indented

  The else branch of the unarchive help check has its content at the same indentation level as the else keyword itself, violating the project's sh-basic-offset: 4 convention. The parallel archive block above it is properly indented.
  Functionally harmless but makes the control flow harder to read.`

atheurer:
I assume we'd be using this to migrate users' results to a shared infra server?

k-rister:
> I assume we'd be using this to migrate users' results to a shared infra server?

Yes.

I'm going to address your review comments above and update this PR.

k-rister:
All review findings addressed:

**1. Unarchive --remote leaves downloaded archive behind** — Fixed in `fd3b222`. Added `rm -f "${ARCHIVE_FILE}"` after successful extraction, matching the local-only unarchive path.

**2. Temp archive deleted on upload failure** — Fixed in `0fd3154`. On failure, the staged archive path is now printed for manual retry instead of being immediately deleted.

**3. Skip-existing check swallows rclone errors** — Fixed in `11eb67d`. Removed `2>/dev/null`, capture the `_run_rclone` exit code, and fail early with a clear error message instead of falling through to the upload.

**4 & 5. Python rclone execution + duplicate arg-building logic** — Acknowledged but deferred. `result-processor.py` already runs inside the controller container (launched by `bin/_main`), so `subprocess.run(["rclone", ...])` works correctly — rclone is available in the container image. The `_build_rclone_args_py` Python implementation duplicates the bash `_build_rclone_args`, but the attempted fix (a standalone bash script or sourcing `bin/_remote_archive` from Python) introduced more complexity than the duplication it eliminated: `_remote_archive`'s functions depend on `_run_rclone` which needs the full podman setup from `bin/base` (session IDs, container args, etc.), so eliminating the Python copy required either duplicating the arg-building logic in a third location (a standalone script) or wiring up the entire `bin/base` sourcing chain from Python — both worse than the current two-copy status quo. Will revisit if/when a new backend type is added.

**6. Unarchive else-block indentation** — Fixed in `f7adc94`. Added 4 spaces of indentation to match `sh-basic-offset: 4`.

atheurer (approved):

PR #635, opened 2026-07-08
feat: overhaul bash tab completions for flag-aware context completion

### LINKED ISSUE #630: Overhaul bash tab completions for consistency and depth
## Summary

The bash completions file (`bin/_crucible_completions`) needs a structural overhaul. The current position-counting approach (`num_words % 2`) doesn't scale to commands with mixed boolean/value flags and variable-length positional arguments. Many commands only have completions at position 3 (the first argument after the command), with no support for deeper positions.

A proper fix likely means switching to a strategy that scans `COMP_WORDS` for recognized flags regardless of position, tracks which flags have already been used, and offers the remaining valid options.

### Structural issues

1. **No completion beyond position 3 for `archive` and `unarchive`.** These commands accept interleaved flags and multiple positional args (`crucible archive dir1 dir2 --remote test-s3 --local`), but completions only fire at position 3. After that, nothing is offered:
   - `crucible archive dir1 <tab>` — no completions (should offer more dirs + flags)
   - `crucible archive --local <tab>` — no completions (should offer dirs + `--remote` + `--delete-local`)
   - `crucible archive --remote test-s3 <tab>` — no completions (should offer dirs + `--local`)

2. **No `--remote` value completion anywhere.** `crucible archive --remote <tab>`, `crucible unarchive --remote <tab>`, and `crucible ls --type archive --remote <tab>` should offer remote backend names from `services.json`.

3. **The `num_words % 2` pattern breaks with boolean flags.** The even/odd heuristic assumes every flag takes exactly one value argument. Commands that mix boolean flags (`--local`, `--delete-local`) and value flags (`--remote <name>`) break this pattern at deeper positions. Pre-dates the archive work — `rm` has the same issue with the optional instance positional argument.

### Missing completions

4. **`get result` / `get metric` — no deeper completions.** `crucible get result --run <tab>` should offer run IDs. `crucible get metric` should offer `--run`, `--source`, `--type`, `--period`, `--begin`, `--end`, `--breakout`, `--filter`.

5. **`opensearch` subcommands with `--name` — no name completion.** `crucible opensearch remove --name <tab>` etc. could complete with instance names from `crucible opensearch list-instance-names`.

6. **`ls --remote` value completion missing.** After `crucible ls --type archive --remote`, the next word should complete with remote names.

7. **`archive config` deeper flag completions.** `crucible archive config add <name> <tab>` should offer `--type --credentials --bucket --tls-verify --setting`. `crucible archive config default <tab>` could offer remote names or `--clear`.

8. **`console` — no position-3 completions.** It's in the top-level word list but has no case in the position-3 switch. Should get file completions like `wrapper`.

### Consistency issues

9. **`help` subcommand list doesn't match all commands with specific help.** The completions only list commands with explicit handlers in `_help`. Commands like `opensearch`, `get`, etc. get no completion even though the general help says "crucible help <command>" works.

10. **Unquoted variable comparisons.** Multiple lines use `${COMP_WORDS[1]} == "..."` without quoting the left side. Should be `"${COMP_WORDS[1]}" == "..."`.

11. **`tags` deeper completions incomplete.** `crucible tags --action add <tab>` should offer `--result-dir` and `--tags`, but only the `--action` value is completed.

## Notes

- Several of these gaps involve archive/unarchive/remote functionality from crucible#629 (not yet merged)
- The fix should be a single structural overhaul, not incremental patches

### LINKED ISSUE #634: result-completion should bypass the logger pipeline
## Summary

`crucible result-completion` is invoked by the bash tab completions (`bin/_crucible_completions`) to dynamically generate run directories, archive names, and run IDs. These calls go through the full `bin/crucible` entry point, which wraps everything in the logger pipeline (`_logger.py`). This has two consequences:

1. **Unnecessary overhead** — the logger setup (pipe creation, background process, SQLite writes) adds latency to every tab completion, even though completion output has no reason to be logged.

2. **Prevents completion status feedback** — writing progress messages to `/dev/tty` during slow completions (e.g., "loading run IDs...") is impossible because the logger also writes to the terminal, causing the two output streams to race and produce garbled output.

## Proposed fix

Give `result-completion` a fast path in `bin/crucible` that skips the logger pipeline and runs `_main` directly, similar to how other lightweight commands could bypass logging. The completion output goes straight to stdout without pipe indirection.

## Context

Discovered while overhauling `bin/_crucible_completions` (issue #630). The completions rewrite adds dynamic completion for many more commands, making the logger overhead more noticeable.

k-rister:
## Summary

- Complete rewrite of `bin/_crucible_completions` — replaces fragile position-counting (`num_words % 2`) with per-command `_complete_<command>` functions that scan `COMP_WORDS` for recognized flags regardless of position
- Dynamic value completion for `--remote`, `--run`, `--result-dir`, `--name` (opensearch instances), and session IDs
- `help` subcommands expanded to all commands; `update` now includes `controller-image`; `start`/`stop` filters already-used services
- All variable expansions properly quoted; all variables declared `local`; removed global `-o filenames` (uses `compopt` only where needed)
- `result-completion` bypassed from logger pipeline — no reason to log tab completion output, and the overhead was significant
- Suppressed spurious error messages during completion mode in `result-processor.py`
- Fixed `bin/repo` `exit` → `return` bug that prevented `crucible repo info crucible` from producing output
- Updated CLAUDE.md to cover CLI parameter modifications/removals (not just additions) and reference the per-command completer pattern

## Test plan

- [ ] `crucible archive dir1 <tab>` — offers more dirs + flags
- [ ] `crucible archive --remote <tab>` — offers remote names
- [ ] `crucible get result --run <tab>` — offers run IDs (no error messages)
- [ ] `crucible get metric <tab>` — offers all metric flags
- [ ] `crucible ls --type archive --remote <tab>` — offers remote names
- [ ] `crucible opensearch remove --name <tab>` — offers `--name`
- [ ] `crucible tags --action add <tab>` — offers `--result-dir --tags`
- [ ] `crucible start opensearch <tab>` — offers remaining services
- [ ] `crucible update <tab>` — includes `controller-image`
- [ ] `crucible help <tab>` — lists all commands
- [ ] `crucible repo info crucible` — produces output (was broken)

Resolves #630
Resolves #634

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -133,7 +133,7 @@ Follow existing code style conventions exactly — check surrounding code for pa
 - **Podman usage**: Use `podman_wrapper()` and related functions from `bin/base`; never call podman directly
 - **JSON processing**: Use `jq_query()` helper from `bin/base`; validate configs against schemas in `schema/`
 - **Variable naming**: Lowercase with underscores for locals; uppercase for exported/environment variables (e.g., `CRUCIBLE_HOME`, `CRUCIBLE_CONTROLLER_IMAGE`)
-- **CLI parameters**: When adding a new CLI parameter, option, or subcommand, update all three locations: `bin/_help` (main help output), `bin/_crucible_completions` (bash tab completions), and the relevant command's own help text
+- **CLI parameters**: When adding, modifying, or removing a CLI parameter, option, or subcommand, update all three locations: `bin/_help` (main help output), `bin/_crucible_completions` (bash tab completions), and the relevant command's own help text. In `_crucible_completions`, each command with deeper completion has its own `_complete_<command>` function — update the flag inventory and value-completion cases in the appropriate function.
 - **Comments**: Write comments that explain the intent behind the code — why a decision was made, what constraint is being satisfied, or what behavior would surprise a future reader. Focus on the WHY rather than restating WHAT the code does.
 
 ## Language Strategy
--- a/bin/_crucible_completions
+++ b/bin/_crucible_completions
@@ -2,159 +2,693 @@
 # -*- mode: sh; indent-tabs-mode: nil; sh-basic-offset: 4 -*-
 # vim: autoindent tabstop=4 shiftwidth=4 expandtab softtabstop=4 filetype=bash
 
+# ---------------------------------------------------------------------------
+# Helper functions: dynamic data retrieval
+# ---------------------------------------------------------------------------
+
+_crucible_get_subproject_names() {
+    find "${CRUCIBLE_HOME}/subprojects/" -type l -printf '%f\n' 2>/dev/null
+}
+
+_crucible_get_subproject_paths() {
+    find "${CRUCIBLE_HOME}/subprojects/" -type l -printf '%P\n' 2>/dev/null
+}
+
+_crucible_get_remote_names() {
+    jq -r '."remote-archive".remotes // {} | keys[]' "${CRUCIBLE_HOME}/config/services.json" 2>/dev/null
+}
+
+_crucible_get_run_dirs() {
+    crucible result-completion --type run-dir 2>/dev/null | tr -d '\r'
+}
+
+_crucible_get_archives() {
+    crucible result-completion --type archive 2>/dev/null | tr -d '\r'
+}
+
+_crucible_get_run_ids() {
+    crucible result-completion --type run-id 2>/dev/null | tr -d '\r'
+}
+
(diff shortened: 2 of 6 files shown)


tc-wilson (approved):

atheurer (approved):
Looks good — the per-command completion functions are a big improvement over the position-counting approach. Clean helper for filtering used flags, proper quoting, and the result-completion pipeline bypass makes sense for tab completion performance. The repo exit→return fix is a nice catch too.

PR #641, opened 2026-07-27
fix: tls-verify=false ignored in remote archive rclone args

k-rister:
## Summary

- The jq `//` (alternative) operator treats both `null` and `false` as falsy, so `."tls-verify" // true` always returned `true` — even when `tls-verify` was explicitly set to `false` in the remote config. This made `--no-check-certificate` unreachable, breaking remotes with self-signed or custom CA certificates.
- Replace with an explicit null check so only a missing field defaults to `true`.

## Test plan

- [ ] Configure a remote with `--tls-verify false` and a self-signed cert endpoint
- [ ] Verify `crucible archive --remote <name> <run>` succeeds (previously failed with TLS errors)
- [ ] Verify a remote without `tls-verify` set still defaults to verifying certificates

### DIFF
--- a/bin/_remote_archive
+++ b/bin/_remote_archive
@@ -43,7 +43,7 @@ function _build_rclone_args() {
     cred_file=$(jq_query ${SERVICES_CFG} --arg name "${remote_name}" '."remote-archive".remotes[$name].credentials')
 
     local tls_verify
-    tls_verify=$(jq_query ${SERVICES_CFG} --arg name "${remote_name}" '."remote-archive".remotes[$name]."tls-verify" // true')
+    tls_verify=$(jq_query ${SERVICES_CFG} --arg name "${remote_name}" '."remote-archive".remotes[$name]."tls-verify" | if . == null then true else . end')
 
     _rclone_args=()
 


atheurer (approved):
## Review

### Correctness

The fix is correct. The replacement expression:

```jq
| if . == null then true else . end
```

correctly distinguishes:
- Key **absent** (jq yields `null`) → defaults to `true` (verify certs)
- Explicitly set to **`false`** → passes through `false` → bash comparison `== "false"` succeeds → `--no-check-certificate` is added
- Explicitly set to **`true`** → passes through `true` → flag omitted

The bash consumer (`[ "${tls_verify}" == "false" ]`) is compatible with jq's string output of boolean values.

### Minor Observations

- **Pre-existing: no validation of the `tls-verify` value.** If someone puts `"tls-verify": "false"` (a JSON string instead of a boolean), jq outputs `"false"` with quotes and the bash comparison silently fails to match. Not introduced by this PR, but worth a follow-up.
- **The other `// empty` uses in the file** (for `endpoint` and `region`) are fine — those are string values that can never be boolean `false`, so the `//` footgun doesn't apply there.
- **Test plan checkboxes are unchecked** — confirm those steps are run before merging.

### Summary

Minimal, correct, and well-described fix for a real jq footgun. Approved.

Issue #654, opened 2026-08-11
Add 'crucible tools list' CLI command for tool/subtool discovery


atheurer:
## Problem

There is no way to programmatically discover what tools crucible has installed, what subtools they support, what parameters they accept, or what metrics they produce. Consumers (including AI agents) must crawl the filesystem and parse bash scripts to discover this information.

## Proposed solution

Add a `crucible tools list` command that walks `subprojects/tools/*/`, reads each tool's metadata files, and outputs structured JSON.

### Depends on

- perftool-incubator/crucible#653 — `tool-metadata.json` and `multiplex.json` must exist in each tool subproject before this command can aggregate them.

### Output format

```bash
$ crucible tools list
```

```json
{
  "tools": [
    {
      "name": "kernel",
      "description": "Linux kernel performance tools — CPU frequency, power, hardware counters, and tracing",
      "subtools": [
        {
          "name": "turbostat",
          "description": "CPU frequency, power, and thermal monitoring via MSR polling",
          "cdm_indexed": false
        },
        {
          "name": "perf",
          "description": "Hardware performance counter sampling",
          "cdm_indexed": false
        }
      ],
      "params": {
        "presets": {"basic": [{"arg": "subtools", "vals": ["turbostat"]}, {"arg": "interval", "vals": ["10"]}]},
        "validations": {"subtools": {"description": "comma-separated list", "args": ["subtools"], "vals": "..."}}
      }
    },
    {
      "name": "sysstat",
      "description": "System activity reporting — mpstat, sar, iostat, pidstat",
      "subtools": [
        {"name": "mpstat", "description": "Per-CPU utilization", "cdm_indexed": true},
        {"name": "sar", "description": "System activity (network, memory, scheduler)", "cdm_indexed": true},
        {"name": "iostat", "description": "Block device I/O statistics", "cdm_indexed": true},
        {"name": "pidstat", "description": "Per-process CPU usage", "cdm_indexed": true}
      ],
      "params": {
        "presets": {"basic": [{"arg": "subtools", "vals": ["mpstat,sar,iostat,pidstat"]}, {"arg": "interval", "vals": ["3"]}]}
      }
    },
    {
      "name": "procstat",
      "description": "Per-process and per-CPU interrupt and softirq statistics",
      "subtools": [],
      "cdm_indexed": true,
      "params": {
        "presets": {"basic": [{"arg": "interval", "vals": ["3"]}]}
      }
    }
  ]
}
```

### Options

```
[...]

Issue #653, opened 2026-08-11
Add machine-readable tool metadata: tool-metadata.json and multiplex.json for all tools


atheurer:
## Problem

Crucible benchmarks have `multiplex.json` for parameter definitions (presets, validations), but **tools have no equivalent metadata**. Tool parameters, subtool declarations, and descriptions exist only as bash comments and `getopt` strings inside `*-start` scripts — not machine-readable.

This makes it impossible for automation (including AI agents) to discover:
- What tools exist and what they do
- Which tools have subtools (and what those subtools are)
- What parameters each tool accepts and their valid values
- Which tools/subtools produce CDM-indexed metrics vs raw file output only

### Current state

| File | What it has | What it's missing |
|------|-------------|-------------------|
| `rickshaw.json` | Tool name, start/stop scripts, endpoint whitelist/blacklist | Description, params, subtools |
| `workshop.json` | Build dependencies (packages, source tarballs) | Everything semantic |
| `*-start` scripts | Params as bash `getopt` longopts; subtool lists as bash comments | Not machine-readable |
| `multiplex.json` | **Does not exist for tools** (only benchmarks) | Everything |

### Tools with subtools today

Discovered by reading bash scripts — this should be in metadata:

- **kernel** → turbostat, perf, speed-select-util, trace-cmd, sysfs-trace (params: `--subtools`, `--interval`, `--record-opts`, `--base-freq`, `--turbo-freq`, `--core-power`, `--trace-cmd-record-opts`, `--sysfs-trace-setup`, `--sysfs-trace-cleanup`)
- **sysstat** → mpstat, sar, iostat, pidstat (params: `--subtools`, `--interval`)
- **ethtool** — no subtools (params: `--interval`, `--interfaces`)
- **forkstat** — no subtools (params: `--events`)
- **ftrace** — no subtools (params: `--record-opts`)
- **mlxreg** — no subtools (params: `--devices`, `--sensors`, `--interval`)
- **nvidia** — no subtools (params: `--interval`)
- **ovs** — no subtools (params: `--interval`)
- **power** — no subtools (params: `--interval`, `--username`, `--password`, `--endpoints`, `--plugin`)
- **procstat** — no subtools (params: `--files`, `--interval`)
- **rt-trace-bpf** — no subtools (params: `--cpu-list`)

### 1. Add `tool-metadata.json` to each tool subproject

New file alongside `rickshaw.json` and `workshop.json`:

```json
{
  "tool": "kernel",
[...]

PR #655, opened 2026-08-12
feat: add tool/benchmark metadata schemas and discovery CLI

k-rister:
## Summary

- Adds `schema/tool-metadata.json` and `schema/benchmark-metadata.json` — machine-readable description and CDM-output-manifest schemas for tool/benchmark subprojects, enforcing mutual exclusivity between the "has subtools/sub-benchmarks" shape and the flat `cdm_indexed` shape via `oneOf`. These land in crucible (not rickshaw) since `crucible tools list`/`crucible benchmarks list` are their only consumers.
- Adds `crucible tools list` and `crucible benchmarks list` (new `bin/list-subprojects.py`, run inside the controller container via `podman_run` like the other `bin/*.py` scripts): aggregates `rickshaw.json` (name fallback), `{tool,benchmark}-metadata.json` (schema-validated in-process, falling back to name-only on failure), and `multiplex.json` into a `table` (default) or `json` view per subproject.
- Wires the new commands into `bin/_main`, `bin/_help`, and `bin/_crucible_completions`.
- Documents the new file formats in `docs/implementing-a-new-tool.md` / `docs/implementing-a-new-benchmark.md`, including the (now-merged, rickshaw PR#865) tool `multiplex.json` backward-compatibility requirement, and adds the new commands to CLAUDE.md's Common Commands.

Addresses perftool-incubator/crucible#653 and #654. Companion to rickshaw PR#865 (merged), which made tool `multiplex.json` functionally live.

## Test plan

- [x] Both schemas pass `jsonschema.Draft7Validator.check_schema`
- [x] All 4 existing reference-implementation metadata files (`tool-kernel`, `tool-sysstat`, `tool-procstat`, `bench-fio`) validate against the new schemas
- [x] `oneOf` mutual-exclusivity confirmed to reject invalid shapes, not just accept valid ones
- [x] `crucible tools list` / `crucible benchmarks list` run live against real subprojects (table and json formats), correct counts (3/14 tools and 1/16 benchmarks with metadata)
- [x] Deliberately-invalid metadata file confirmed to warn and fall back gracefully instead of crashing the listing
- [x] `crucible help` shows the new commands; new bash-completion functions exercised directly


k-rister:
## PR Review: crucible#655 — feat: add tool/benchmark metadata schemas and discovery CLI

**Summary:** Adds `schema/tool-metadata.json` and `schema/benchmark-metadata.json` (with `oneOf`-enforced subtools/flat-shape mutual exclusivity) plus a new `bin/list-subprojects.py` aggregator, wired into `crucible tools list` / `crucible benchmarks list`.
**Changed files:** 9

### Bugs
- **[bin/list-subprojects.py:162] A `description` field containing an embedded newline corrupts the table output's column alignment.** Confirmed by reproduction: a `tool-metadata.json` with `"description": "Line one\nLine two continues here"` renders as two misaligned lines instead of one row, throwing off every column after it. `truncate()` only checks total length, not for embedded newlines, so even short multi-line descriptions slip through untouched.

### Issues
- **[schema/tool-metadata.json:56, schema/benchmark-metadata.json (equivalent)] The documented `cdm_sources` "present when cdm_indexed is true" invariant is not schema-enforced.** Tested directly against the schema: `{"cdm_indexed": true}` with no `cdm_sources`, and `{"cdm_indexed": false, "cdm_sources": [...]}`, both validate successfully despite the field description explicitly stating the conditional relationship. `crucible tools list` would then show `CDM Indexed: yes` for an entry with nothing to look up. An `if`/`then` (available in the Draft7 dialect this schema targets, per the PR's own test plan) would close this.
- **[bin/list-subprojects.py] No unit tests added for the new aggregation/formatting logic**, despite `bin/test_manage_opensearch.py` existing as a precedent for testing `bin/*.py` scripts in this repo. `build_entry()`'s three-way fallback (missing file / JSON-parse error / schema-validation error) and the tools-vs-benchmarks `subgroup_field` dispatch are exactly the kind of branching logic that regresses silently without coverage. Not blocking — crucible's own CI doesn't yet have the test-gate wired up (unlike toolbox/rickshaw) — but worth flagging.

### Documentation
- **[bin/_help:19] Help text lists "params" as part of `tools list`/`benchmarks list` output, but the default table view has no params column** — only `--format json` surfaces the aggregated `multiplex.json` presets/validations. A user going by `crucible help` alone could reasonably expect params in the default table.

### File Coverage
- [x] CLAUDE.md — No issues found
- [x] bin/_crucible_completions — No issues found
- [x] bin/_help — 1 doc note (params wording)
- [x] bin/_main — No issues found
- [x] bin/list-subprojects.py — 1 bug, 1 issue (missing tests)
- [x] docs/implementing-a-new-benchmark.md — No issues found
- [x] docs/implementing-a-new-tool.md — No issues found (verified factual claims: `EC_EMPTY_SET_FAIL=6` in multiplex.py, `apply_tool_multiplex()`'s role/id-dropping behavior in rickshaw-run.py, `VALID_CLASSES`/`VALID_AGGREGATIONS` in toolbox's cdm_metrics.py — all accurate)
- [x] schema/benchmark-metadata.json — 1 issue (cdm_indexed/cdm_sources not enforced)
- [x] schema/tool-metadata.json — 1 issue (same, shared root cause)

Missing from diff: none identified — checked for a central schema registry (none exists; each schema is referenced directly by its consumer, consistent with existing convention), checked `spec/` (correctly out of scope — that directory documents crucible's own root-level configs, not per-subproject files), and grepped `docs/` for stale references to the new concepts (none found).

### Limitations
- Did not execute `crucible tools list`/`crucible benchmarks list` through the actual `podman_run` container path — verified `bin/list-subprojects.py`'s logic directly on the host (with `jsonschema` installed, matching the version already baked into the controller image via rickshaw/toolbox/multiplex `workshop.json`), which exercises identical code paths since the script has no other container-specific dependencies.
- Did not interactively test bash tab-completion; verified the added completion functions by tracing `COMP_WORDS`/`num_index` logic by hand.
- Relied on the PR's stated live-testing test plan for the actual `crucible tools list` / `crucible benchmarks list` container invocation end-to-end.

### Verdict

**Request changes** — the table-formatting bug (embedded-newline description) is a confirmed, easily-triggered defect in the PR's primary new user-facing output; it's a small fix (e.g. collapse whitespace/newlines before rendering) but should land before merge. The schema-enforcement gap and missing tests are good follow-ups but not blocking on their own.

k-rister:
Addressed in 6ad669f:

- **Bug (blocking):** fixed — `print_table()` now collapses embedded whitespace/newlines (`" ".join(description.split())`) before truncating. Reproduced the exact repro case live before and after; `--format json` still returns the raw, unmodified description.
- **Schema gap:** fixed — added `if`/`then`/`else` (Draft7) to both `tool-metadata.json` and `benchmark-metadata.json`, at the top level and inside each subtool/sub-benchmark item, so `cdm_indexed: true` requires `cdm_sources` and `cdm_indexed: false` forbids it. Verified both invalid directions now rejected and both valid shapes still accepted; all 4 existing reference-implementation metadata files still validate unchanged.
- **Missing tests:** addressed — added `bin/test_list_subprojects.py` (10 cases) covering `build_entry()`'s fallback branches (missing metadata, invalid JSON, schema-validation failure, multiplex.json handling), the tool/benchmark `subgroup_field` dispatch, and a regression test locking in the newline fix.
- **Doc wording:** fixed — `crucible help` now notes the default table view doesn't include `multiplex.json` params, only `--format json` does.

Nothing skipped.

k-rister:
## PR Review: crucible#655 — feat: add tool/benchmark metadata schemas and discovery CLI

**Summary:** Adds `schema/tool-metadata.json` / `schema/benchmark-metadata.json` (oneOf-enforced description + CDM-output-manifest schemas) and a new `crucible tools list` / `crucible benchmarks list` CLI (`bin/list-subprojects.py`) that aggregates `rickshaw.json`, the new metadata file, and `multiplex.json` per subproject.

**Changed files:** 10

### Bugs

- **[bin/list-subprojects.py:102] `multiplex.json` that parses to valid JSON but isn't an object crashes the entire listing** — If any tool or benchmark subproject's `multiplex.json` is syntactically valid JSON whose top-level value isn't an object (e.g. `[]`, `null`, or a bare string — an easy hand-authoring mistake, since `multiplex.json` is explicitly documented in this PR as something tool/benchmark authors write by hand), `multiplex.get("presets", {})` raises an uncaught `AttributeError` (`'list' object has no attribute 'get'`). Only `json.JSONDecodeError` is caught (line 105), so this isn't a syntax error — it slips through. Reproduced directly against the shipped script: a `multiplex.json` containing `["not", "an", "object"]` crashes `build_entry()`, which propagates up and kills `crucible tools list`/`crucible benchmarks list` for *every* subproject, not just the offending one. This directly contradicts the PR's own stated design goal (docstring: "a malformed file for one subproject should not prevent the rest of the listing from working") and test-plan claim about graceful fallback — that claim was verified only for the metadata file (which is protected by schema `"type": "object"` validation), not for `multiplex.json` (which has no such guard).

- **[bin/list-subprojects.py:64] `rickshaw.json` that parses to valid JSON but isn't an object crashes the entire listing** — Same root cause, different call site: `rickshaw_data[config["rickshaw_field"]]` assumes `rickshaw_data` is a dict. If `rickshaw.json` is valid JSON but not an object (e.g. a bare list), this raises `TypeError: list indices must be integers or slices, not str`, which is not in the `except (json.JSONDecodeError, KeyError)` clause (line 65) and propagates uncaught, crashing the whole command the same way. Reproduced directly: a `rickshaw.json` containing `["not", "an", "object"]` crashes `build_entry()`.

### File Coverage

- [x] CLAUDE.md — No issues found
- [x] bin/_crucible_completions — No issues found (verified `_complete_subprojects_list`, top-level word list ordering, and dispatch wiring)
- [x] bin/_help — No issues found (help text matches implemented behavior, verified live)
- [x] bin/_main — No issues found (dispatch logic exercised live: default `list`, explicit `list`, unsupported-argument error path all behave as expected)
- [x] bin/list-subprojects.py — 2 bugs
- [x] bin/test_list_subprojects.py — No issues found; all 10 tests pass (does not cover the non-dict-JSON crash scenario above, which is why it slipped through)
- [x] docs/implementing-a-new-benchmark.md — No issues found
- [x] docs/implementing-a-new-tool.md — No issues found (cross-checked the `multiplex.json`/`apply_tool_multiplex`/`EC_EMPTY_SET_FAIL` claims against the actual rickshaw source — accurate)
- [x] schema/benchmark-metadata.json — No issues found (`oneOf`/`if`-`then`-`else` mutual-exclusivity empirically verified with 6 valid/invalid probe documents)
- [x] schema/tool-metadata.json — No issues found (same empirical verification)

### Limitations

This review verified schema enforcement and the CLI's happy-path/fallback behavior by running `crucible tools list`/`crucible benchmarks list` live (table and JSON formats, `--name` filter) and by running the new unit test suite. The two crash bugs were reproduced by invoking `build_entry()` directly against the on-disk script with the same Python/jsonschema environment used by the container, not through a full `podman run` round-trip — I'm confident the code path is identical, but did not independently confirm the container's `jsonschema` package version matches. I did not test behavior with a very large number of subprojects, nor with real (non-reference) upstream benchmark/tool repos' metadata files.

### Verdict

**Request changes** — the `multiplex.json`/`rickshaw.json` non-object-JSON crash bugs are the blocking issue: they defeat the per-subproject fallback isolation this PR explicitly designed for and tested, and either bug can take down `crucible tools list`/`crucible benchmarks list` entirely from a single subproject's hand-authored file mistake.

k-rister:
Addressed in 2be9fc4:

- **Both bugs (blocking):** fixed with the same root-cause fix — `rickshaw.json`/`multiplex.json` that parse as valid JSON but aren't objects now raise a `TypeError` that's caught alongside the existing `json.JSONDecodeError`/`KeyError` handling, falling back gracefully instead of propagating uncaught.
- Added 3 regression tests (13 total, all passing): non-dict `rickshaw.json`, non-dict `multiplex.json`, and one that directly locks in the design property that broke — a malformed subproject no longer prevents every other subproject from appearing in `collect_entries()`.
- Verified live through the actual `podman_run` container path (not just unit tests): a real `crashtest-bad` subproject with `multiplex.json` containing `["not", "an", "object"]` alongside a good subproject now warns to stderr and both still appear in `crucible tools list --format json`, exit code 0.

Nothing skipped.

atheurer (commented):
## Review

Overall well-structured PR. Good test coverage, sensible fallback behavior, and thorough documentation. A few issues worth discussing before merging.

---

### Issues

**1. Schema version is an exact regex match — `schema/tool-metadata.json`, `schema/benchmark-metadata.json`**

```json
"pattern": "^2026\\.08\\.11$"
```

Any schema evolution will cause existing metadata files to fail validation and silently fall back to name-only, even if the new schema is backward-compatible. A looser pattern like `^\d{4}\.\d{2}\.\d{2}$` or just removing the version constraint and relying on `additionalProperties: false` on the envelope object would be safer. If strict versioning is the intent, the degradation-to-name-only fallback means breakage won't surface loudly.

**2. Unquoted `$@` in `bin/_main`**

```bash
list_cmd="${CRUCIBLE_HOME}/bin/list-subprojects.py --kind ${subproject_kind} $@"
${podman_run} ... ${CRUCIBLE_CONTROLLER_IMAGE} ${list_cmd}
```

Both `list_cmd` and `${CRUCIBLE_CONTROLLER_IMAGE}` are unquoted on expansion. This follows the existing `_main` style so it won't regress anything, but `--name` values with spaces would break. Low risk in practice since subproject names don't have spaces.

**3. No guard for `CRUCIBLE_HOME` unset — `bin/list-subprojects.py`**

```python
CRUCIBLE_HOME = os.environ.get("CRUCIBLE_HOME")
```

If unset, `os.path.join(None, ...)` in `collect_entries` raises a bare `TypeError`. Since this always runs inside the controller container where the env var is set, it won't matter in practice, but a quick `if not CRUCIBLE_HOME: sys.exit(...)` near the top would give a cleaner error for debugging.

**4. Silent skip vs. warning inconsistency — `bin/list-subprojects.py`**

Missing `rickshaw.json` returns `None` silently; a malformed one emits a warning then returns `None`. In a directory that *should* have a `rickshaw.json`, the silent case could obscure misconfigured installs.

**5. Tests use simplified schemas — `bin/test_list_subprojects.py`**

The test `TOOL_SCHEMA`/`BENCHMARK_SCHEMA` are minimal stubs that don't include `subtools`, `cdm_sources`, `output_files`, `oneOf`, or `if/then/else`. The schema's main correctness invariant — mutual exclusivity between the subtools shape and the flat shape — isn't exercised by any test. A test that validates a known-good metadata file against the *actual* schema file would close this gap.

---

### Minor observations

- `subgroup_singular = subgroup_field.rstrip("s")` works correctly here (`"subtools"` → `"subtool"`, `"sub-benchmarks"` → `"sub-benchmark"`), but `rstrip` strips *all* trailing matching characters, not just one. `[:-1]` would be more explicit.
- The `make_subproject` helper in tests has a long inline conditional for choosing the metadata filename — a local variable would improve readability.

---

Items 1 and 5 are the most worth addressing before merge: the version pinning creates a schema evolution trap, and the missing schema-level tests leave the `oneOf` logic unverified. Items 2–4 are low severity given the runtime context.

atheurer (dismissed):

k-rister:
Thanks for the review, @atheurer. Addressed in 48b6a75:

**Fixed:**
- **Item 4 (silent-skip vs. warning inconsistency):** missing `rickshaw.json` now warns to stderr and skips, matching every other anomaly-handling path in this file.
- **Item 5 (tests don't exercise oneOf/if-then-else):** real gap — added `TestRealToolMetadataSchema`/`TestRealBenchmarkMetadataSchema` (13 new tests) that load the actual shipped `schema/tool-metadata.json`/`schema/benchmark-metadata.json` and validate the exact valid/invalid shapes (subtools-shape, flat-shape, both-present, neither-present, cdm_indexed/cdm_sources mismatches). 26 tests total, all passing.
- **Minor:** `rstrip("s")` → `[:-1]` (more precise, as suggested); extracted the test helper's inline metadata-filename conditional into a local variable.

**Deliberately left as-is:**
- **Item 1 (exact-match schema version regex):** checked `rickshaw/schema/tool.json`/`benchmark.json` — both have used an identical exact-match pattern (`^2020\.03\.18$`) unchanged since 2020. These two new schemas match that established precedent rather than deviating from it. The "fall back on any validation failure" behavior is also an existing, previously-reviewed design choice for listing resilience generally, not something specific to version pinning.
- **Item 2 (unquoted `$@`):** your own comment already notes it matches existing `_main` style and is low-risk — not changing it.
- **Item 3 (no `CRUCIBLE_HOME` guard):** this script only ever runs via `podman_run`, which always sets `CRUCIBLE_HOME` explicitly. Adding a guard here would be validating a scenario that can't actually occur given the real invocation path.

atheurer (commented):
## Follow-up review

Good progress from the previous rounds — the crash bugs, schema enforcement, doc wording, and `rstrip("s")` fix are all addressed. Two remaining items:

---

**1. Schema version pinning still present — `schema/tool-metadata.json`, `schema/benchmark-metadata.json`**

```json
"pattern": "^2026\\.08\\.11$"
```

This was in the first review and didn't get picked up in the subsequent two. Any future schema evolution will silently downgrade every existing metadata file to name-only output unless authors also update the version string in their files. A date-shaped pattern like `^\d{4}\.\d{2}\.\d{2}$` would be more future-friendly without giving up version tracking.

---

**2. `TestRealBenchmarkMetadataSchema` has 5 tests vs. `TestRealToolMetadataSchema`'s 8 — `bin/test_list_subprojects.py`**

Three cases present for tools are absent for benchmarks:

- `test_flat_shape_cdm_indexed_false_is_valid` — the common case of a non-CDM benchmark is untested as a positive
- `test_neither_sub_benchmarks_nor_cdm_indexed_is_invalid` — no test that a document missing both shapes is rejected (tools has `test_neither_subtools_nor_cdm_indexed_is_invalid`)
- `test_sub_benchmark_cdm_indexed_true_without_sources_is_invalid` — the per-item `if/then/else` for `sub-benchmarks` items isn't exercised (tools has `test_subtool_cdm_indexed_true_without_sources_is_invalid`)

The schemas are structurally identical in these areas so the gaps look like oversights. The third missing test in particular exercises the per-item constraint that was one of the fixes from the previous round.

---

Everything else looks clean: the missing-`rickshaw.json` warning, the `make_subproject` helper cleanup, and both real-schema test classes are solid additions. Ready to approve modulo the two items above.

k-rister:
Addressed in cc4bf93. On the version-pinning item specifically: my previous fix (a loose \`^\d{4}\.\d{2}\.\d{2}\$\` regex) overcorrected — it accepted any date-shaped string forever, which loses the ability to invalidate old versions on a breaking change entirely.

The right fix is \`enum\` listing the currently-supported version(s), not a pattern. Surveyed every schema's version field in the repo to confirm this is the established, actively-used convention rather than a one-off: \`workshop/schema.json\` has accumulated 7 enum entries over 5 years, \`rickshaw-run.json\` has 2 — both grow the enum on compatible changes and drop stale entries on incompatible ones. \`tool.json\`/\`benchmark.json\` (which the original exact-match pattern was modeled on) turned out to be single-version outliers that have never needed a second version, which is why they look identical under `pattern` or `enum` — the wrong precedent to generalize from.

Both schemas now use \`"enum": ["2026.08.11"]\`, with the description spelling out the append-on-compatible/replace-on-breaking convention. Verified: the 4 real reference files still validate, a garbage version string is rejected, and — the actual point — a well-formed but unlisted/unsupported version is also correctly rejected now.

Also closed the \`TestRealBenchmarkMetadataSchema\` coverage gap from the follow-up review (3 missing tests added, now matching \`TestRealToolMetadataSchema\`'s 8 cases 1:1). 29 tests total, all passing.

atheurer (approved):
All issues from prior rounds are addressed. The `enum` approach for version versioning is the right call — matches the `workshop/schema.json` convention and the description clearly explains the append-on-compatible/replace-on-breaking contract. `TestRealBenchmarkMetadataSchema` now has full 8-case parity with `TestRealToolMetadataSchema`. The two deliberately-deferred items (unquoted `$@` in `_main`, no `CRUCIBLE_HOME` guard) have solid justifications. LGTM.

PR #656, opened 2026-08-13
feat: gate crucible-ci on crucible's own unit tests

### LINKED ISSUE #637: ci: add unittest workflow for bin/test_manage_opensearch.py
bin/test_manage_opensearch.py is a 335-line Python unittest suite with 25+ test methods covering manage_opensearch.py functionality (load/save JSON, add/remove/update instances, CLI args). It uses unittest.mock and requires no external dependencies beyond the standard library.

No GitHub Actions workflow currently runs these tests.

**Action:** Add a unittest.yaml workflow that runs pytest (or python -m unittest) against this file on PRs.

Jira: PERFNFV-399
Epic: PERFNFV-398

k-rister:
## Summary

- Adds a `call-unittest` job to `crucible-ci.yaml` that gates the expensive `call-real-core-release-crucible-ci` reusable-workflow call behind crucible's own unit tests (`bin/test_list_subprojects.py` + `bin/test_manage_opensearch.py`, 63 tests total, previously unwired to any CI).
- New `.github/workflows/unittest.yaml` (`workflow_call`/`workflow_dispatch` only — no `pull_request` trigger, so tests don't run twice per PR).
- Same 3-part wiring already merged in toolbox (#130) and rickshaw (#864): `call-unittest` shares `call-real-core-release-crucible-ci`'s `if:` condition, is added to that job's `needs:`, and is *also* added directly to `crucible-ci-complete`'s `needs:` — a failed gate produces `skipped` (not `failure`) on the downstream job, and `crucible-ci-complete`'s check only looks for `failure`/`cancelled`, so without this direct dependency a failing gate would silently pass the required check.
- No branch-protection change needed: crucible's ruleset only requires `crucible-ci-complete` for this path (verified via `gh api .../rulesets/3637247`, not the `.github/rulesets/*` backup), and `call-unittest` now feeds that job directly.
- Updates `docs/how-ci-works.md` with a new "Local test gating" section (parallel to "Docs-only filtering"), the workflow hierarchy diagram, and the Level 1 trigger description — this pattern wasn't documented anywhere yet despite being live in two other repos already.

Closes #637 — `bin/test_manage_opensearch.py` now runs on every PR via this gate.

## Test plan

- [x] `bash -n`/YAML-parsed both workflow files
- [x] Ran all 63 tests in a fresh venv with only `jsonschema` installed (matching the CI environment exactly) — all pass
- [x] Confirmed `unittest.yaml` is not in `crucible-ci.yaml`'s docs-only skip-file list (it's the test payload, not orchestration)
- [ ] Confirm on this PR's own CI run: `call-unittest` passes before `call-real-core-release-crucible-ci` starts, `call-faux-core-release-crucible-ci` still correctly skips for non-docs changes

### DIFF
--- a/.github/workflows/crucible-ci.yaml
+++ b/.github/workflows/crucible-ci.yaml
@@ -43,8 +43,13 @@ jobs:
       - name: Display changes
         run: echo '${{ toJSON(steps.filter.outputs) }}' | jq .
 
-  call-real-core-release-crucible-ci:
+  call-unittest:
     needs: changes
+    if: ${{ github.event.pull_request.head.repo.fork != true && (github.event_name == 'workflow_dispatch' || needs.changes.outputs.only-docs != 'true') }}
+    uses: ./.github/workflows/unittest.yaml
+
+  call-real-core-release-crucible-ci:
+    needs: [ changes, call-unittest ]
     if: ${{ github.event.pull_request.head.repo.fork != true && (github.event_name == 'workflow_dispatch' || needs.changes.outputs.only-docs != 'true') }}
     uses: perftool-incubator/crucible-ci/.github/workflows/core-release-crucible-ci.yaml@main
     with:
@@ -61,7 +66,7 @@ jobs:
     uses: perftool-incubator/crucible-ci/.github/workflows/faux-core-release-crucible-ci.yaml@main
 
   crucible-ci-complete:
-    needs: [ call-real-core-release-crucible-ci, call-faux-core-release-crucible-ci ]
+    needs: [ call-unittest, call-real-core-release-crucible-ci, call-faux-core-release-crucible-ci ]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -1,0 +1,27 @@
+name: unittest
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  unittest:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          python3 -m venv .venv
+          source .venv/bin/activate
+          pip install jsonschema
+
+      - name: Run crucible python unit tests
+        run: |
+          source .venv/bin/activate
+          cd bin
+          python3 -m unittest discover -s . -p "test_*.py" -v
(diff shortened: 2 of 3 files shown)


atheurer (dismissed):
I have reviewed this PR and approve the changes. They are solid, correct, and well-documented.

Here are a few minor optimizations and suggestions you might want to apply (I've verified these locally and they work perfectly):

### 1. Remove Unused Dependency (`jsonschema`)
Crucible's top-level tests (under `bin/`) do not import or use `jsonschema` (though some subprojects do). We can safely remove `pip install jsonschema` from `.github/workflows/unittest.yaml` to save a bit of build time.

### 2. Simplify Test Execution Path
The test runner can be invoked directly from the root of the repository by specifying the source directory:
```yaml
      - name: Run crucible python unit tests
        run: |
          source .venv/bin/activate
          python3 -m unittest discover -s bin -p "test_*.py" -v
```
This is a bit cleaner as it avoids the need to `cd bin`.

### 3. Run Unit Tests on Fork PRs (High Value)
Since unit tests run on public, GitHub-hosted `ubuntu-latest` runners and do not require any repository secrets, we can remove the fork exclusion from the `call-unittest` job in `crucible-ci.yaml`:
```yaml
  call-unittest:
    needs: changes
    if: ${{ github.event_name == 'workflow_dispatch' || needs.changes.outputs.only-docs != 'true' }}
    uses: ./.github/workflows/unittest.yaml
```
This allows fork PRs to get fast, automatic unit-test feedback without exposing any self-hosted runners or secrets (since `call-real-core-release-crucible-ci` remains securely gated and skipped on forks).

---
Awesome work! Happy to approve this.

k-rister:
Thanks for the review! Applied #2 in e6a1b5e. Pushing back on the other two:

**#1 (remove jsonschema)** — this is actually needed. `bin/list-subprojects.py:8` does `from jsonschema import validate, ValidationError`, and `test_list_subprojects.py` imports that module directly. Without `jsonschema` installed, the whole `unittest discover` run fails at import time before any test even executes, taking `test_manage_opensearch.py`'s tests down with it. Verified locally: removing the pip install reproduces the crash.

**#3 (run on fork PRs)** — agree it's technically safe (no secrets, `ubuntu-latest`), but leaving `call-unittest` off the fork exclusion. The gate is deliberately kept on the *same* `if:` condition as the job it gates (`call-real-core-release-crucible-ci`) across every repo that's adopted this pattern (toolbox, rickshaw, multiplex) — if the gate's scope diverges from what it's gating, you can end up with a gate that runs in cases the thing it protects never does (or vice versa), which is a correctness trap we've hit before in this CI work. Fork PRs get closed by `fork-check.yaml` anyway, so there's no real feedback loss.

atheurer (approved):
Review completed. The architecture is clean, avoids standard GHA dependency pitfalls, and successfully gates the integration tests behind the fast Python unit test suite. Verified the tests pass locally in ~0.4s.

PR #659, opened 2026-08-17
fix: start CDM server when OpenSearch is already running

### LINKED ISSUE #651: bug: start_opensearch does not call start_cdm_server when OpenSearch is already running
## Problem

When `start_opensearch` detects OpenSearch is already running, it returns without calling `start_cdm_server`. The CDM server may not be running, and `npm ci` in `start-server.sh` is never executed, so `node_modules` is never installed in the cdmq directory.

`index_run` then calls `add-run.sh` and gets:
```
Error: node_modules not found. Run start-server.sh or 'npm install' in .../cdmq first.
ERROR: Could not index result into OpenSearch [rc=1]
```

The benchmark succeeds but results are not indexed.

## Root cause

In `bin/base`, `start_cdm_server` is only called inside the branch that starts OpenSearch fresh:

```bash
if ! podman_running crucible-opensearch; then
    # ... starts opensearch, calls start_cdm_server ...
else
    echo "...appears to be running"  # CDM server never checked or started
fi
```

## Fix

After detecting OpenSearch is running, still check and start the CDM server if absent:

```bash
else
    echo "...appears to be running"
    echo -n "Checking for CDM server"
    if ! podman_running crucible-cdm-server; then
        start_cdm_server
    fi
fi
```

## Reproduction

1. Run a benchmark — OpenSearch + CDM server start, node_modules installed
2. Run another benchmark without stopping services
3. `start_opensearch` sees OpenSearch running, skips `start_cdm_server`
4. CDM server absent → `add-run.sh` fails with node_modules not found

k-rister:
## Summary
- `start_opensearch()` only called `start_cdm_server` inside the branch that starts OpenSearch fresh. If OpenSearch was already running (e.g. left running from a prior session), the CDM server was never checked or started.
- Result: the benchmark succeeds, but `index_run` later fails at `add-run.sh` with a confusing `Error: node_modules not found. Run start-server.sh or 'npm install' in .../cdmq first.` -- far removed from the actual root cause.
- `start_cdm_server()` is already idempotent (no-ops if its container is already running), so the fix just calls it unconditionally in the "OpenSearch already running" branch too, matching the existing branch's pattern of propagating a CDM server startup failure through `start_opensearch()`'s own return code.

Fixes #651

## Test plan
- [x] `bash -n bin/base`
- [x] Live reproduction: stopped only `crucible-cdm-server` (`--rm`, so this fully removes it) while leaving `crucible-opensearch` running, then ran `crucible start opensearch` -- confirmed it now detects the missing CDM server and recreates it (`Checking for CDM server...not present` -> `Successfully started CDM server on port 3000`)
- [x] Confirmed idempotency: running `crucible start opensearch` again with both already up reports `...appears to be running` for both and does nothing further

### DIFF
--- a/bin/base
+++ b/bin/base
@@ -512,6 +512,18 @@ function start_opensearch() {
                 fi
             fi
         fi
+    else
+        # OpenSearch was already running -- still need to check the CDM
+        # server companion service, since it's tracked as a separate
+        # container and may have been stopped (or never started)
+        # independently of OpenSearch. start_cdm_server() is a no-op if
+        # it's already running.
+        start_cdm_server
+        local CDM_RC=$?
+        if [ ${CDM_RC} != 0 ]; then
+            echo "WARNING: OpenSearch already running but CDM server failed to start"
+            RC=${CDM_RC}
+        fi
     fi
     return ${RC}
 }


atheurer (approved):
### Code Review: Pull Request #659 (Start CDM Server when OpenSearch is Already Running)

I have performed a thorough review of this PR and tested it locally. The change is elegant, correct, and perfectly resolves the service initialization issue (#651).

#### Key Review Findings:
* **Root Cause Correctly Addressed**: It correctly ensures the `crucible-cdm-server` companion service is checked and started when `crucible-opensearch` was already running (the `else` branch of `start_opensearch()`).
* **Safe and Idempotent**: Since `start_cdm_server` is idempotent (returning 0 immediately if the container is already running), adding this check has no negative performance impact or side-effects.
* **Standards and Scoping**: Code styling, 4-space indentation, and variable scoping (declaring `local CDM_RC` and properly propagating return code `RC`) conform to Crucible's bash development conventions.

This PR is ready to merge. **Approved!**

PR #674, opened 2026-09-09
docs: clarify remotehosts SSH address selection

### LINKED ISSUE #673: docs: define controller-to-remote SSH address selection for remotehosts
## Problem\n\nThe remotehosts run-file uses remotes[].config.host for Crucible controller-to-remote SSH, file transfer, and engine orchestration. Current documentation does not clearly distinguish this control-plane address from benchmark dataplane addresses. In multi-network environments, especially public clouds, the canonical hostname or dataplane interface may not be reachable for SSH from the controller.\n\n## Required guidance\n\nUpdate docs/how-endpoints-work.md to explain that:\n\n- remotes[].config.host may be a hostname or IP address; the format is not the selection rule.\n- The selected address must be verified as SSH-reachable from the Crucible controller using the configured user/key.\n- The controller-to-remote access network is independent from the benchmark dataplane network and benchmark-selected interfaces.\n- When multiple networks are available, prefer an access/management path for controller SSH so the dataplane remains dedicated to the workload.\n- The selected address must be used consistently for endpoint validation, deployment, file transfer, and cleanup.\n- Benchmark-specific mechanisms (for example, uperf ifname/service discovery) determine dataplane addressing separately.\n\n## Evidence\n\nAn agent generated a valid uperf run file with 192.168.50.1 and 192.168.50.2 as remotehosts. Those addresses were correct for the uperf dataplane, but the Crucible controller could not SSH to them, so remotehosts validation failed before execution. The controller could SSH to the endpoint hostnames.\n\nRelated benchmark-specific guidance: perftool-incubator/bench-uperf#93.

atheurer:
Closes #673\n\nDocument how to select remotes[].config.host for controller-to-remote SSH:\n\n- Clarify that hostnames and IP addresses are both valid, but reachability from the controller is the criterion.\n- Explain that a non-dataplane management path is preferred when available, but the dataplane path may be the only usable route because of network policies or firewalls.\n- Require testing SSH access to the exact selected address with the configured user and key.\n- Distinguish controller SSH addressing from benchmark-specific dataplane discovery.


k-rister:
This PR won't need to run CI since it's documentation only so lets see if we can get it through even though it is from a fork.

k-rister:
Apparently not, our protection is too good for that.

PR #675, opened 2026-09-09
feat: add MCP server interface for Crucible

### LINKED ISSUE #668: Request: provide an MCP server interface for Crucible
## Summary

Create and maintain an MCP (Model Context Protocol) server interface for Crucible so agentic tools can interact with Crucible through a native, discoverable, and safety-conscious interface.

A primary motivating use case is [agentic-perf](https://github.com/atheurer/agentic-perf), but the interface should be designed for other agentic clients as well.

## Motivation

Today, an agent that wants to use Crucible must understand and invoke its CLI, construct JSON run files, locate available benchmarks and tools, monitor execution, and retrieve/index results through command-specific conventions. An MCP server could expose these capabilities as typed tools and resources, allowing an agent to discover the installed Crucible capabilities and use them with less shell-specific glue and fewer opportunities for malformed requests.

## Proposed scope

- Define the supported MCP transport and deployment model (for example, a local stdio server and/or a remotely hosted server).
- Expose safe discovery operations for Crucible capabilities, including benchmarks, tools, endpoints, schemas, and relevant configuration.
- Expose run lifecycle operations: validate or construct a run request, start a run, inspect status, stream or retrieve logs, and cancel a run where supported.
- Expose result operations: list runs, inspect run metadata, retrieve metrics/artifacts, and index results where appropriate.
- Reuse Crucible validation and authorization boundaries rather than duplicating business logic in the MCP layer.
- Define error handling, timeouts, cancellation, concurrency, and potentially long-running operation behavior.
- Document how agentic clients such as agentic-perf authenticate, discover capabilities, and perform a complete run/query workflow.
- Add automated tests and an example client or integration test.

## Design questions

- Which operations should be read-only by default, and how should execution/destructive actions require explicit confirmation?
- Should the server invoke the existing CLI, call shared Bash functions, or introduce a stable internal API beneath both interfaces?
- What result representations are most useful to agents without exposing excessively large payloads?
- How should remote Crucible hosts, credentials, multi-user isolation, and auditability be handled?
- Which MCP resources/prompts, if any, should accompany tools?

## Initial acceptance criteria

- A documented MCP server can be started against a Crucible installation.
- An MCP client can discover available Crucible capabilities and schemas.
- An MCP client can validate and launch a representative run, observe its lifecycle, and retrieve a useful result/metric summary.
- Unsafe or destructive operations have explicit safeguards and documented behavior.
- The interface has versioning/compatibility guidance and automated coverage.

This issue is intended to track the design and implementation effort; implementation can be split into follow-up issues once the initial architecture is agreed.

— AI-signed: Codex | model: GPT-5 | effort: not specified

k-rister:
## Summary

Implements the Crucible MCP server and client-facing tool interface for benchmark discovery, run validation and execution, result/period/metric queries, and logger inspection.

The implementation includes durable asynchronous job tracking, idempotent submissions, input-path policy enforcement, audit logging, structured JSON-RPC errors, service lifecycle integration, configuration migration, and controller-image dependency support.

## Validation

- MCP test suite: 36 passed, 3 skipped
- Bash syntax and diff checks passed
- Review fixes committed independently throughout the branch

Fixes #668

— AI-signed: Codex | model: GPT-5 | effort: not specified


atheurer (changes requested):
I found several blocking issues in the current implementation:

1. **[P1] Streamable HTTP notification handling is not compliant.** Every POST returns HTTP 200 with a JSON body. A standard `notifications/initialized` notification falls through to “method not found,” but accepted Streamable HTTP notifications must return HTTP 202 with no body. `GET /mcp` also returns 404 rather than the required SSE response or 405. Strict MCP clients may reject this transport. See: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports

2. **[P1] Restart reconciliation permanently strands active jobs.** `RunManager.reconcile()` moves jobs to `recovery_required` or `unknown_after_crash`, then permanently skips those states on future reconciliation. Both states remain in `ACTIVE_JOB_STATES`, so `service_control()` subsequently refuses every service shutdown. There is no MCP or CLI operation that resolves either state. Reconciliation needs to inspect the process identity, event/run directory, containers, and result metadata and transition jobs to an appropriate terminal state, or provide an explicit operator-resolution command.

3. **[P1] Crucible/CDM identifiers are never captured.** The schema exposes `rickshaw_run_id`, `cdm_run_id`, and `runner_container_id`, but the event producer/consumer only persists lifecycle state and `run_directory`. Those identifier fields therefore remain null for real runs, preventing reliable correlation between an MCP job and subsequent CDM metric queries.

4. **[P1] CDM readiness failures are misclassified as benchmark failures.** Any nonzero `crucible run` exit—including failure of the final result-summary query after successful execution/indexing—is persisted as lifecycle state `failed`. The configured readiness timeout only waits for a local summary file when a client calls `get_run_summary`; it does not retry CDM after indexing or eventually transition result status to `unavailable`. This loses the intended distinction between workload failure and CDM consistency timeout.

5. **[P2] Job and audit files are not created with the documented private permissions.** `start_mcp_server()` uses plain `mkdir -p` and `touch` for the database and audit log, so the normal root umask creates mode 0755 directories and 0644 files. The implementation plan requires the job database to be root-owned mode 0600; the audit data should be protected similarly.

6. **[P2] The HTTP endpoint does not validate `Origin`.** The Streamable HTTP specification requires Origin validation to prevent DNS-rebinding attacks. This is especially important because an authenticated request can launch privileged Crucible workloads.

Validation: the focused MCP suite passes locally (36 passed, 3 live-integration checks skipped), and current CI is green. The missing coverage aligns with the findings: there is no standards-based MCP-client handshake test, execution/restart end-to-end test, or live launch → index → summary test.

k-rister:
Current planned path after this PR:

1. Keep this PR focused on the MCP foundation and complete CI/live validation without adding the run-planning subsystem.
2. Merge the current MCP implementation once CI and live smoke testing are satisfactory.
3. Develop the planning capability as a separate cross-repository effort:
   - expose a reusable bounded expansion API in multiplex;
   - add the canonical RunPlanner and plan model in rickshaw;
   - add thin prepare_run and estimate_run MCP adapters in Crucible.
4. Keep rickshaw authoritative for run semantics; Crucible should provide policy, transport, response bounds, and redaction rather than duplicate expansion logic.
5. Do not rebuild the controller image for ordinary source changes. The existing workflow should trigger a rebuild only when dependency metadata such as workshop.json changes.

The planning work will remain separate so it does not expand this PR’s review and debugging surface.

— AI-signed: Codex | model: GPT-5 | effort: not specified

atheurer (changes requested):
Request changes

1. `mcp/crucible_mcp/operations.py:2270` — `_load_run_metadata()` does not enforce the metadata nesting limit. On Python 3.14, a 10,000-level JSON document parses successfully, so `test_load_run_metadata_maps_embedded_json_recursion_error` fails instead of returning `result_too_large`. Validate nesting explicitly after parsing rather than depending on `RecursionError`.

2. `mcp/crucible_mcp/operations.py:536` — `list_local_runs` permits up to 1,000 entries while returning unbounded tag content. Each metadata file can be 1 MiB and `_validated_tags()` has no per-tag or aggregate bound, allowing a response approaching 1 GiB and requiring repeated JSON serialization. Add a cumulative response-size limit/pagination and indicate truncation.

Validation: `PYTHONPATH=mcp python3 -m unittest discover -s mcp/tests -v` — 163 tests run, 1 failed, 3 skipped.

atheurer (approved):
Follow-up review: the two previously requested changes are addressed.

- `_load_run_metadata()` now validates metadata nesting explicitly.
- `list_local_runs` now bounds serialized responses, supports offsets, and reports truncation.

Revalidated with `PYTHONPATH=mcp python3 -m unittest discover -s mcp/tests -p "test_*.py" -v`: 169 passed, 3 skipped.

